PR #1, opened 2020-11-20
New condensed version of Swift BAT Position notice.

TeresaSheets:


lpsinger:
Nice! This is a pretty thorough model of the existing GCN schema for Swift as an AVRO packet. Does it parse?

swift_bat_position.json
```diff
@@ -0,0 +1,30 @@
+{"namespace": "gcn.notices",
+ "type": "record",
+ "name": "SWIFT_BAT_GRB_POSITION",
+ "fields": [
+     {"name": "TITLE",         "type": "string"},
+     {"name": "NOTICE_DATE",   "type": "string",          "doc": "date and time of notice creation"},
+     {"name": "NOTICE_TYPE",   "type": "string",          "doc": "type of notice"},
+     {"name": "TRIGGER_NUM",   "type": ["int", "string"], "doc": "integer or string of digits uniquely identifying event" },  
+     {"name": "SEG_NUM",       "type" : "long",           "doc": "Swift serial number (for Swift notice types only)" },
+     {"name": "GRB_RA_J2000",  "type": "double",          "doc": "ra of GRB/Transient in J2000 epoch [deg]"},
```
lpsinger, line 10: * J2000 has been the preferred equinox for a long time. Do we need to specify that in the field name?
* How about just "ra", so that the field name is general to all types of transients, not just GRBs?

jracusin: This notice type is specifically only for GRBs, I think BAT would generate a different notice type if it was a non-GRB source.  Though for simplicity and comparison to other notice types, and the general use of RA/Dec, it would be easier to parse these streams if not GRB specific.

lpsinger:
Note that Avro allows you to set [aliases](https://avro.apache.org/docs/current/spec.html#Aliases) for fields. This might be an option to provide new fields with consistent lower-case capitalization while retaining the old names too.

ra-dec-positions.json
```diff
@@ -0,0 +1,12 @@
+    { "type": "record",
```
lpsinger, line 1: This record is not used.

astronomical-configurations.json
```diff
@@ -0,0 +1,39 @@
+    {"type": "record",
```
lpsinger, line 1: This record is not used.

ra-dec-positions.json
```diff
@@ -0,0 +1,12 @@
...
+     "namespace": "gcn.notices",
+     "name": "ra-dec-positions",
+     "fields": [
+                {"name": "grb-ra-j2000",    "type": "double", "doc": "ra of GRB/Transient in J2000 epoch [deg]"},
+                {"name": "grb-ra-current",  "type": "double", "doc": "ra of GRB/Transient in Current epoch [deg]"},
+                {"name": "grb-ra-1950",     "type": "double", "doc": "ra of GRB/Transient in 1950 epoch [deg]"},
```
lpsinger, line 7: I really do not think that we need to provide precomputed coordinates in epochs other than J2000 (including true of date).

ra-dec-positions.json
TeresaSheets: This is the referenced schema that would be used in the future enhanced versions.

astronomical-configurations.json
TeresaSheets: This is the referenced schema that would be used in the future enhanced versions.

PR #28, opened 2023-02-27
Just validation stuff

dakota002:


dakota002:
The $id fields need to be sorted out, I will make a separate PR if Vidushi hasn't gotten to it already

lpsinger (changes requested):

validate.mjs
```diff
@@ -1,25 +1,28 @@
...
+  const schemas = await Promise.all(
+    filesInDir.map(async (file) =>
+      JSON.parse(
+        await readFile(file, {
+          encoding: 'utf-8',
+        })
+      )
+    )
+  )
+  ajv.addSchema(schemas).compile(true)
```
lpsinger, line 24: What happens on error?

Can we please retain the original behavior that:

* if there are no errors, the process exits cleanly
* if there are any validation errors, they are all printed with the format `${path}: ${e.message}`, and then the process exits with status 1
* if there are any errors that are not validation errors, then they are re-thrown

validate.mjs
```diff
@@ -1,25 +1,28 @@
...
+  return true
```
lpsinger, line 25: No need for a return value:

```suggestion
```

lpsinger (changes requested):

validate.mjs
```diff
@@ -1,17 +1,28 @@
...
-    return true
+    const schemas = await Promise.all(
+      filesInDir.map(async (file) =>
+        JSON.parse(
+          await readFile(file, {
+            encoding: 'utf-8',
+          })
+        )
+      )
+    )
```
lpsinger, line 24: Move out of the `try` block.

lpsinger (changes requested):

Vidushi-GitHub (approved):

lpsinger (approved):

Issue #44, opened 2023-03-15
Move icecube schema out of AMON


blaufuss:
We will start sending second copies  (and eventually more feature filled) of the IceCube Astrotrack Notices via GCN Kafka.  In GCN classic, these are sent via AMON, but will not be going forward.  

Any concerns if I move these out of icecube/amon to icecube/ ?

lpsinger:
Fine with me.

PR #46, opened 2023-03-29
Move things out of amon subdirectory. 

blaufuss:
All new alerts from IceCube won't go through AMON.  First class "icecube" alerts.


blaufuss:
will close #44

PR #55, opened 2023-04-29
Added a second energy range.

shb46:
This is a change in the HardnessRatio schema. It used to have only one energy range. This commit adds a second one and names both appropriately.


jracusin:
Addressed by #61

PR #57, opened 2023-04-29
Loc galactic

shb46:
Added necessary fields for burst location in galactic, instrument and spacecraft coordinates to the `Localization` schema.


gcn/notices/core/Localization.schema.json
```diff
@@ -34,6 +30,46 @@
...
+      "description": "Galactic longitude [deg]"
+    },
+    "galactic_b": {
+      "type": ["number", "null"],
+      "description": "Galactic latitude [deg]"
+    },
+    "galactic_semimajor_angle": {
+      "type": ["number", "null"],
+      "description": "Position angle of semi-major axis in galactic coordinates [deg]"
+    },
```
lpsinger, line 48: I would rather not include galactic coordinates as we can easily calculate them from other frames (such as ICRS). I understand the convenience of calculating them for users, but we need not include them in the alerts themselves in order to do that.

gcn/notices/core/Localization.schema.json
```diff
@@ -34,6 +30,46 @@
...
+      "description": "Spacecraft phi [deg]"
+    },
+    "spacecraft_theta": {
+      "type": ["number", "null"],
+      "description": "Spacecraft theta [deg]"
+    },
+    "spacecraft_semimajor_angle": {
+      "type": ["number", "null"],
+      "description": "Position angle of semi-major axis in Spacecraft coordinates [deg]"
+    },
```
lpsinger, line 72: I am not sure that these fields belong in the location record, which describes the location of the burst, not the location of the spacecraft.

lpsinger (changes requested):

gcn/notices/core/Localization.schema.json
shb46: You mentioned that at the meeting, then Judy said, it is a useful piece of info to have at hand. I put it in, because it might help with your team's decision.

gcn/notices/core/Localization.schema.json
shb46: These are burst coordinates in the frames of the instrument and the spacecraft respectively. The location of the spacecraft with respect to Earth is in the GeoLoc scheata, which I put in a separate pull.

gcn/notices/core/Localization.schema.json
jracusin: I don't have a strong opinion either way on this.

lpsinger: I do. It's easily derived from ICRS coordinates. Let's omit the Galactic coordinates in order to decrease the surface area of the schema that missions have to implement. We can easily provide Galactic coordinates and many other coordinate frames through APIs and libraries.

shb46 (commented):
Removed fields related to galactic coordinates as requested.

gcn/notices/core/Localization.schema.json
jracusin: please remove the spacecraft phi, theta, semi_major.  Phi and Theta seem to have different definitions for different instruments.  These should be part of the instrument specific schema (e.g. glowbug), with a more detailed description of what the angle is and how it is referenced.

shb46:
If other changes have been requested please point them out.

dakota002:
@shb46 , @jracusin  requested some changes  in a comment above:
>please remove the spacecraft phi, theta, semi_major. Phi and Theta seem to have different definitions for different instruments. These should be part of the instrument specific schema (e.g. glowbug), with a more detailed description of what the angle is and how it is referenced.

shb46:
Thanks for pointing that to me. Is there anything else Im missing?

dakota002:
I think that is all, looks good

PR #75, opened 2023-05-10
Refactors the detectors to be properties

### LINKED ISSUE #74: Array vs. 15 named properties?
https://github.com/nasa-gcn/gcn-schema/blob/fcf288bd8c418c915f291492f452d44e10648be4/gcn/notices/glowbug/alert.example.json#LL24C1-L39C1

Hi,

What was the reason to change the array of 15 with 15 named properties? 

In the current schema the names can be anything? Shouldn't we change that?

Thanks!

dakota002:
Closes #74

### DIFF
--- a/gcn/notices/glowbug/alert.example.json
+++ b/gcn/notices/glowbug/alert.example.json
@@ -20,23 +20,21 @@
   "p_astro": 0.9,
   "classification": { "BNS": 0.9, "NSBH": 0.05, "BBH": 0.05 },
 
-  "detector_status": {
-    "CsI0": "on",
-    "CsI1": "on",
-    "CsI2": "on",
-    "CsI3": "on",
-    "CsI4": "triggered",
-    "CsI5": "on",
-    "CsI6": "on",
-    "CsI7": "on",
-    "CsI8": "off",
-    "CsI9": "on",
-    "CsI10": "on",
-    "CsI11": "on",
-    "CLLB0": "off",
-    "CLLB1": "off",
-    "SSA": "on"
-  },
+  "CsI0": "on",
+  "CsI1": "on",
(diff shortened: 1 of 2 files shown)


dakota002:
@shb46 is this what you had in mind for separating out each detector?

shb46:
@dakota002: Yes. Except that I didn't mean to make anything required.

dakota002:
No problem, I can remove the required part

shb46:
@dakota002: Thanks!

Vidushi-GitHub:
@shb46 for both human-reading and data-base prospects writing each detector status as separate field (would appear as separate 15 columns table for Glowbug only) doesn't seem convenient way. I had a chat with @jracusin, we would prefer to see it as a single field type. Like Fermi also report as single field such as detector="00001001110". Is it fine if we go with previous way, detector status  = {CsI0=on, CsI1=off...etc}? @dakota002

shb46:
@Vidushi-GitHub (CC: @dakota002, @jracusin, @lpsinger): Either way is ok, since it is the same information presented in a slightly different way. Correct me if I'm wrong, but are you proposing a string, as in:

> `"detector_status":"CsI0=on, CsI1=off,`...`,SAA=on}"`.

It is more concise and easily parsable, for those who need these data. At the very beginning I was thinking of using the Fermi-GBM style. Enhanced to show 3-states per detector, it could look like:

> `"detector_status":"1111 1211 0111 00 1"`

Also, a reminder that an array of 15 string enums was one of the previous suggestions:

> `"detector_status":["on","on","on","on","on","triggered","on","on","off","on","on","on","off","off","on"]`

Since you know best what you need for a best GCN system, please pick and tweak and keep me posted.

PR #79, opened 2023-05-13
GUANO notices

Tohuvavohu:
Make GUANO notices work with the new flat structure and core schema
Fix Alert.schema.json to have consistent property names

### DIFF
--- a/gcn/notices/swift/bat/guano/alert.schema.json
+++ b/gcn/notices/swift/bat/guano/alert.schema.json
@@ -1,26 +1,42 @@
 {
   "$id": "https://gcn.nasa.gov/schema/gcn/notices/swift/bat/guano/alert.schema.json",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Swift/BAT-GUANO Alert",
+  "description": "Candidate gamma-ray transient reported from the BAT-GUANO targeted search",
   "type": "object",
-  "properties": {
-    "alert_reporter": {
-      "$ref": "/schema/gcn/notices/core/Reporter.schema.json"
-    },
-    "alert_info": {
+  "allOf": [
+    {
       "$ref": "/schema/gcn/notices/core/Alert.schema.json"
     },
-    "event": {
+    {
       "$ref": "/schema/gcn/notices/core/Event.schema.json"
     },
-    "datetime": {
+    {
       "$ref": "/schema/gcn/notices/core/DateTime.schema.json"
     },
-    "localization": {
+    {
       "$ref": "/schema/gcn/notices/core/Localization.schema.json"
     },
-    "statistics": {
+    {
       "$ref": "/schema/gcn/notices/core/Statistics.schema.json"
+    },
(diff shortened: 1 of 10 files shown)


gcn/notices/swift/bat/guano/alert.schema.json
```diff
@@ -1,26 +1,40 @@
...
+    },
+    {
+      "$ref": "/schema/gcn/notices/core/Reporter.schema.json"
+    },
+    {
+      "$ref": "/schema/gcn/notices/core/AdditionalInfo.schema.json"
+    }
+  ],
+  "properties": {
+    "follow_up_event": {
```
Vidushi-GitHub, line 31: This field can be referred to core-schema at https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/core/Event.schema.json
Such that in example can be provided as:
"id": [bn....] or alternate
"id_source": [Fermi]

gcn/notices/swift/bat/guano/alert.schema.json
```diff
@@ -1,26 +1,40 @@
...
+      "type": "string",
+      "description": "Name or trigger time of the external trigger that launched the search."
+    },
+    "follow_up_type": {
```
Vidushi-GitHub, line 35: Similarly, we have a field for this one in the Reporter.schema as messenger. Could you please replace it with already existing fields in the core?

Tohuvavohu: This is inadequate, because I already use messenger:EM for the GRB notice, and need a way of denoting the messeger of the source trigger.

gcn/notices/swift/bat/guano/alert.schema.json
Tohuvavohu: The core schema you reference is inadequate because it does not communicate the event that triggered the followup

gcn/notices/swift/bat/guano/alert.schema.json
Vidushi-GitHub: Ok, but then name of event will cover GRB.

Tohuvavohu: How do you figure?

Also, consider, GRB is just one example usage 
 For example S190814 is pretty opaque to event type.

Tohuvavohu:
Note that this PR now changes part of the core schema, as described in the naming inconsistency at the top of this issue https://github.com/nasa-gcn/gcn-schema/issues/82

Tohuvavohu:
Despite the broken validator, I am mildly confident that the things in this PR are indeed correct and can be merged @dakota002

Tohuvavohu:
Added 
$schema:true 
to properties field and
unevaluatedProperties:false

in order to force the validator to actually check all fields, and not fail on $schema.
As agreed here https://github.com/nasa-gcn/gcn-schema/issues/80

Ready to go.

Tohuvavohu:
Rebased.

PR #86, opened 2023-05-17
Renames some example files to use a consistent pattern

dakota002:
In order to make the schema browser understand which example files correspond to the appropriate schema files I need some kind of naming convention. 

The pattern will be `<Name>[.<Optional Specification>].example.json`


dakota002:
@Tohuvavohu  and @blaufuss I made a small change to the naming of some of the example files you provided. I am working on the schema browser for gcn.nasa.gov and this consistency will make it easier to programmatically determine which examples correspond to their respective schemas, without having to load all the examples and read their schema fields. Are you both cool with this change?

blaufuss:
works for me

Tohuvavohu:
LGTM

PR #90, opened 2023-05-19
correct reference

Vidushi-GitHub:
ref word is changed to reference


dakota002 (approved):

dakota002:
@jracusin Just wanted to make sure you see this before it is merged in. I recommended this change because  `$ref` is a keyword and having another property as `ref` will definitely lead to confusion.

jracusin (approved):

PR #92, opened 2023-05-21
Initial template of IGWN notices from table

dakota002:
Rough draft based on table at https://emfollow.docs.ligo.org/userguide/content.html#


dakota002:
@Vidushi-GitHub @jracusin This could probably be refined to utilize the existing core schema objects, but its a start. 
Its important to note that the alert types in these notices are all caps. I am looking into what needs to be added in the validator to allow the `pattern` keyword to be used, which would allow us to define a RegEx which is case insensitive, but at the moment this passes

lpsinger:
We should just use the official schema: https://git.ligo.org/emfollow/igwn-gwalert-schema

PR #95, opened 2023-05-23
anyOfs changed to allOf, small update to schema to fix type problem.

dakota002:
The change from anyOf to allOf fixes the false positives in validation. 

As a note, omitting fields from the objects added with allOf ( example, omitting `alert_type` when using a `$ref` to the core `Alert` schema) is still valid, as these fields are not required.


lpsinger (approved):

lpsinger:
@dakota002, please resolve the merge conflict.

blaufuss (approved):

blaufuss (approved):

lpsinger:
Yes, it has all of the review approvals.

PR #96, opened 2023-05-23
Removes fermi folder, will be added when ready

dakota002:
Are there others that need to be removed in this PR?


lpsinger (approved):

Vidushi-GitHub:
We don't want Fermi folder in tag version, is there any way to keep it as a reference, may be branch?

PR #98, opened 2023-05-23
Allow any key value pair for reference

dakota002:
Description explains the usage for referencing notices/circulars. Please make any suggestions though if this is not what the intended idea was

### DIFF
--- a/gcn/notices/core/FollowUp.schema.json
+++ b/gcn/notices/core/FollowUp.schema.json
@@ -19,8 +19,7 @@
     },
     "reference": {
       "type": "object",
-      "additionalProperties": { "type": "number" },
-      "description": "Reference as distributed by the notices or circulars or ATel, ex. gcn.notices.Fermi.gbm.alert: 4642"
+      "description": "Key-value pairs to reference notices, circulars, ATel, etc. For notices, the key should be the topic as a string, and the value should be the timestamp of the notice (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z'). For circulars the key should be 'gcn.circulars' and the value should be the circular ID (ex: 'gcn.circular':32517)"
     }
   }
 }
--- a/gcn/notices/core/FollowUp.schema.json
+++ b/gcn/notices/core/FollowUp.schema.json
@@ -19,7 +19,7 @@
     },
     "reference": {
       "type": "object",
-      "description": "Key-value pairs to reference notices, circulars, ATel, etc. For notices, the key should be the topic as a string, and the value should be the timestamp of the notice (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z'). For circulars the key should be 'gcn.circulars' and the value should be the circular ID (ex: 'gcn.circular':32517)"
+      "description": "Key-value pairs to reference Notices, Circulars, ATel, etc. that point to the reference localization. For notices, provide the topic and timestamp (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z') or topic and alert type with use of ref_ID field (ex: 'igwn.gwalert':'PRELIMINARY'). For Circulars or ATels, the key-value pair should be the system and ID (ex: 'circular':32517 or 'atel':12345)."
     }
   }
 }


lpsinger (commented):
Personally, I think that it would be better not to even advertise a core reference field, until we have thought much more about how to implement it.

dakota002:
Okay, think we should just scrap the field and let it be defined on a per notice basis?

lpsinger:
That's my opinion, yes.

blaufuss:
If there's no clear database backed reference that we can include, let's skip.  For our LVK alert followups, we
do want to be sure to include additional information on what version of the alert we used, since the skymap can change.

To do this, we're planning on encoding this in the 'ref_ID' item from the FollowUp Schema such as:
"ref_ID": "MS230427r-1-Preliminary"

If there are other ideas out there, welcome to hear them

gcn/notices/core/FollowUp.schema.json
```diff
@@ -19,8 +19,7 @@
     },
     "reference": {
       "type": "object",
-      "additionalProperties": { "type": "number" },
-      "description": "Reference as distributed by the notices or circulars or ATel, ex. gcn.notices.Fermi.gbm.alert: 4642"
+      "description": "Key-value pairs to reference notices, circulars, ATel, etc. For notices, the key should be the topic as a string, and the value should be the timestamp of the notice (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z'). For circulars the key should be 'gcn.circulars' and the value should be the circular ID (ex: 'gcn.circular':32517)"
```
Vidushi-GitHub, line 22:
```suggestion
      "description": "Key-value pairs to reference notices, circulars, ATel, etc. For notices, the key should be the topic as a string, and the value should be the timestamp of the notice (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z'). For circulars the key should be 'gcn.circulars' and the value should be the circular ID (ex: 'gcn.circular':32517), similarly for ATel."
```

Vidushi-GitHub:
@jracusin do we wanna keep reference, seems Leo would like to avoid it?

gcn/notices/core/FollowUp.schema.json
jracusin:
```suggestion
      "description": "Key-value pair to reference notices, circulars, ATel, etc. that point to localization being followed up. For notices, the key can be the topic as a string with the value as the timestamp of the notice (ex: 'gcn.notices.glowbug.alert':'2023-01-01T03:24:36.0Z') or another string descriptor of the follow-up source. For Circulars or ATels, the key-value pair should be the system and ID (ex: 'circular':32517 or 'atel':12345)."
```

blaufuss: similar, but I think we were using fields intended for the alert producer.  At the end of the day, we're in production now with:

"ref_ID": "MS230427r-1-Preliminary" type references.  which encodes all we need.  IT does not provide a good clean handle to link back to the original LVK notice, without some trolling on gracedb site.

dakota002: I see, so your setup should still work and this `reference`  update will just allow for any additional generic key-value pairs that might be useful in other cases

blaufuss: correct.  I do think eventually there needs be a simple way to have a reference to notices from others that's easy to find, and makes tracking down the original alert possible.  I guess this will be more important for some notice visualization tool or website so it can make these connections.

lpsinger:
I'd really like to see a design document presenting several different reference designs and a trade study of them. I don't think that any of the reference designs that have been proposed are ready for prime time.

dakota002:
> I'd really like to see a design document presenting several different reference designs and a trade study of them. I don't think that any of the reference designs that have been proposed are ready for prime time.

I think thats fine, but then lets make that a new issue and close this out. The scope of this PR just removes the typing requirement and expands on the description.

lpsinger (commented):
I want to see a design document describing cross-references and presenting a trade study of different options before I approve anything related to them.

dakota002:
Closing this PR, as this issue needs further investigation

Issue #10, opened 2022-02-14
Design record type for spacecraft position and attitude


lpsinger:
Design an Avro record schema to model position and attitude of a satellite in low-Earth orbit (i.e., spacecraft state vector). Draw upon FITS headers for housekeeping data used across multiple HEASARC-supported missions.

jracusin:
This probably also needs pointing info.

Vidushi-GitHub:
First part: Design record schema to model position and attitude of a satellite in low-Earth orbit (i.e., spacecraft state vector) has been implemented in [#56.](https://github.com/nasa-gcn/gcn-schema/pull/56)

PR #99, opened 2023-05-25
Renamed Orbit to Location schema

Vidushi-GitHub:
As discussed, it represents location of spacecraft rather than orbit. https://github.com/nasa-gcn/gcn-schema/issues/10 is still incomplete. @shb46


lpsinger:
This would be easily confused with Localization.

What this data structure really describes is the local geomagnetic field.

Vidushi-GitHub:
I agree. What about SpacecraftLocation?

lpsinger:
Well, it's not the spacecraft location either. It's a random mix of spacecraft location and geomagnetic parameters. I don't think that they belong together.

jracusin:
What about GeoLocation, which is where Boyan started?

gcn/notices/core/Location.schema.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://gcn.nasa.gov/schema/gcn/notices/core/GeoLocBase.schema.json",
+  "$id": "https://gcn.nasa.gov/schema/gcn/notices/core/Location.schema.json",
   "$schema": "https://json-schema.org/draft/2020-12/schema",
   "title": "Spacecraft Geolocation and McIlwain L",
   "description": "Enables GCN providers to include spacecraft geolocation and McIlwain L-Parameter.",
```
jracusin, line 5:
```suggestion
  "description": "Enables GCN providers to include observatory geolocation and McIlwain L-Parameter.",
```

lpsinger:
> What about GeoLocation? This is where Boyan started.

No, that doesn’t accurately describe this data structure either.

I see the use of a data structure that captures the local geomagnetic field AND a separate data structure that captures the full kinematic state of the spacecraft. However, I see no reason why these should be part of the same data structure.

Also, we should seek domain expertise (space weather and orbital dynamics) in the design of both data structures.

jracusin:
@shb46, does Glowbug calculate the McIlwain-L onboard?  Does this value need to be reported?  GBM does not report it in their GCN Classic Notices.  We also found that the GBM Data Tools has a function to calculate McIlwain-L based upon latitude and longitude. https://fermi.gsfc.nasa.gov/ssc/data/analysis/rmfit/gbm_data_tools/gdt-docs/api/api-utils.html

lpsinger:
It's highly unlikely that Glowbug calculates that onboard.

I'm not arguing against a data structure for geomagnetic field parameters. However, I strongly urge you to look do a review of different parametrization and consult some space weather experts on the design.

Vidushi-GitHub:
As no instrument sends notices with spacecraft location and we don't have design idea, thus closing it. We can open a new PR if such requirement arises in future.

PR #101, opened 2023-06-07
Tag script

dakota002:


dakota002:
> The CI script needs to actually commit the changes. Also, if the CI script is triggered by tag events, then it's actually running too late: the tagged commit needs to have the version recorded in the JSON files.

Ahh right, converting this to a draft while I work on the update

dakota002:
Okay, with this new implementation, whoever is going to do a new version update only has to run a single command:
`npm run update-tags --tag=vX.X.X`

This triggers the following chain of events:
- Parse out tag value from --tag
  - If the tag is missing or doesn't follow the vX.X.X[optional additional content] pattern, exit the process
- Run prettier
- Checks out a new branch `Update-${tag}`, adds all files, commits and pushes 

I am looking into having the PR get created automatically once the branch is pushed

The workflow now only triggers when the PR gets merged. Once merged, the main branch gets tagged to `vX.X.X`, parsed from the PR name

dakota002:
Now it generates a PR, cmd line usage is in my previous comment.
On merge of this resulting PR, the main branch gets tagged. 
I think the logic makes sense, but let me know if this is wrong/over-engineered
I also updated the readme to give some guidance on how to use it

lpsinger (changes requested):
- [ ] Trigger the workflow by changes on the main branch that alter the version number in package.json.
- [ ] The `$id`, `$ref`, and `$schema` fields should be updated on the tagged version, but not on the main branch. You can accomplish this either by updating the version numbers on the main branch, committing and tagging, and changing them back; or by simply changing the version numbers and tagging a dangling commit.

dakota002:
> * [ ]  Trigger the workflow by changes on the main branch that alter the version number in package.json.
> * [ ]  The `$id`, `$ref`, and `$schema` fields should be updated on the tagged version, but not on the main branch. You can accomplish this either by updating the version numbers on the main branch, committing and tagging, and changing them back; or by simply changing the version numbers and tagging a dangling commit.

Cool, working on updating it now!

dakota002:
I found these after a quick search:
https://github.com/marketplace/actions/detect-and-tag-new-version
https://github.com/marketplace/actions/auto-tag

The first one actually looks like it would handle this case exactly. I can test it out and I'll make an update to this PR if it seems effective

dakota002:
@lpsinger I updated the tag script to use the detect and tag new version action. It simplifies our script a lot, but we still have to hook into the version-command input to run the updates on the schema and id fields and get the commit in there that will actually be tagged. Also its important that the commit is run with the -q flag because the version-command takes whatever is output as the tag value. 

I will have another update for the script that actually writes the file paths once nasa-gcn/gcn.nasa.gov/pull/1010 gets merged

.github/workflows/tag.yml
```diff
@@ -0,0 +1,33 @@
+name: Update From Package
+on:
+  pull_request:
+    types:
+      - 'closed'
+    paths:
+      - 'package.json'
```
lpsinger, line 7: How about just on push to the main branch?

gcn/notices/swift/bat/guano/alert.schema.json
```diff
@@ -1,30 +1,30 @@
...
-      "$ref": "/schema/gcn/notices/core/Statistics.schema.json"
+      "$ref": "/schema/main/gcn/notices/core/Statistics.schema.json"
     },
     {
-      "$ref": "/schema/gcn/notices/core/Reporter.schema.json"
+      "$ref": "/schema/main/gcn/notices/core/Reporter.schema.json"
     },
     {
-      "$ref": "/schema/gcn/notices/core/AdditionalInfo.schema.json"
+      "$ref": "/schema/main/gcn/notices/core/AdditionalInfo.schema.json"
```
lpsinger, line 27: Use relative paths for all of these.

update-tags.mjs
```diff
@@ -0,0 +1,65 @@
+import { readFile, writeFile } from 'fs/promises'
+import { execSync } from 'child_process'
+import { glob as baseGlob } from 'glob'
+
+const version = JSON.parse(await readFile('package.json')).version
+if (!/[0-9]+\.[0-9]+\.[0-9]+.*/.test(version)) {
```
lpsinger, line 6: Why do we need to validate the version number? Why not just take its value verbatim, whatever it is?

update-tags.mjs
```diff
@@ -0,0 +1,65 @@
+import { readFile, writeFile } from 'fs/promises'
+import { execSync } from 'child_process'
```
lpsinger, line 2: You're already using promises. Why not execute external processes asynchronously too?

lpsinger (changes requested):
- Strictly speaking, we don't want to replace _all_ occurrences of `oldValue` with `newValue` --- only the first one. We have already checked if the key `startsWith` the `oldValue`, so we can construct the new value by slicing and concatenating.
- If we push the tagged schema to the main branch rather than creating a dangling tag, then there will be a lot of large commits with low information content on the main branch. What are the pros and cons of keeping all the commits on the main branch vs. making dangling tags?
- We want _releases_ in addition to tags.

update-tags.mjs
lpsinger: Yeah, just remove it. The version number `0.1` would fail validation, for example.

dakota002:
> * Strictly speaking, we don't want to replace _all_ occurrences of `oldValue` with `newValue` --- only the first one. We have already checked if the key `startsWith` the `oldValue`, so we can construct the new value by slicing and concatenating.

The string replace function only should replace the first instance of a match. 
image


> * If we push the tagged schema to the main branch rather than creating a dangling tag, then there will be a lot of large commits with low information content on the main branch. What are the pros and cons of keeping all the commits on the main branch vs. making dangling tags?

My concern at first was that the garbage collector thing would remove the dangling commit, but I read that if the commit is tagged, the gc will not remove it. I'm definitely down to implement it this way, is there a best practice or pattern for making the dangling tag? I haven't (intentionally) made use of them before.

> * We want _releases_ in addition to tags.

I'll look for some existing actions that can do this automatically, pretty sure I came across a few in my previous search

lpsinger:
Would you please make a separate PR with the updates to all of the hardcoded paths in the `.schema.json` and `.json` files? That's just a good idea and I'd like to get that on the main branch immediately, and keep this PR with the tag script small.

dakota002:
Yup! https://github.com/nasa-gcn/gcn-schema/pull/103 should do it!

lpsinger:
Let's try out the following alternative:

- Use [`npm version`](https://docs.npmjs.com/cli/v9/commands/npm-version) for the developer to bump the release on their own system
- `package.json` looks something like this:
```json
{
  "name": "@nasa-gcn/schema",
  "version": "0.0.01",
  "scripts": {
    "version": "node update-version.js && git add -u",
    "postversion": "node update-version.js main && git add -u && git commit -m \"Back to development\""
  }
}
```
- `update-version.js` looks something like this:
```ts
import {version} from 'package.json'

async function updatePaths(ref: string) {
    // Update paths in .json and .schema.json files
}

const [_, ref] = process.argv.slice(2)

await updatePaths(ref || version)
```

PR #104, opened 2023-07-05
Alt tag method

dakota002:
Alternate method as discussed in #101 

@lpsinger , the only concern I have is that this will create the tag/release on the user's fork. Would the user who triggers the update need to be working in a clone of the nasa-gcn repo specifically for this method to work as intended?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,15 @@
-name: Create New Release
+name: Main
+
 on:
   push:
     tags:
       - '*'
 
 jobs:
-  tag-new-versions:
+  build:
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
     steps:
-      - uses: actions/checkout@v3
-      - uses: ncipollo/release-action@v1
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Release
+        uses: softprops/action-gh-release@v1
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,14 @@
+name: Create New Release
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  tag-new-versions:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ncipollo/release-action@v1
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ Your pipeline will generate JSON files following these schema and send alerts to
 
 ## How To Release
 
-New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command. The following steps describe the process for incrementing versions.
+1. Clone the parent repo (nasa-gcn/gcn-schema).
 
-1. Clone the parent repo (nasa-gcn/gcn-schema). This should not be run in your own forked version of this repo.
+   `git clone git@github.com:nasa-gcn/gcn-schema.git `
 
 2. Run `npm version [ major | minor | patch ]`
    This command will handle the intermediate steps of updating, committing, and pushing the path changes in each file as defined in the `version` and `postversion` npm scripts
--- a/README.md
+++ b/README.md
@@ -23,4 +23,4 @@ Your pipeline will generate JSON files following these schema and send alerts to
 
 4.  Push the commit and tag:
 
-    git push && git push origin <tag name>
+        git push && git push origin <tag name>
--- a/README.md
+++ b/README.md
@@ -11,15 +11,16 @@ Your pipeline will generate JSON files following these schema and send alerts to
 
 ## How To Release
 
-1. Clone the parent repo (nasa-gcn/gcn-schema).
+1.  Clone the parent repo (nasa-gcn/gcn-schema).
 
-   `git clone git@github.com:nasa-gcn/gcn-schema.git `
+        git clone git@github.com:nasa-gcn/gcn-schema.git
 
-2. Run `npm version [ major | minor | patch ]`
-   This command will handle the intermediate steps of updating and committing the path changes in each file as defined in the `version` and `postversion` npm scripts
+2.  Run `npm version [ major | minor | patch ]`
 
-3. Review the changes with `git log -p` to make sure that each file is appropriately updated.
+    This command will handle the intermediate steps of updating and committing the path changes in each file as defined in the `version` and `postversion` npm scripts
 
-4. Push the commit and tag:
+3.  Review the changes with `git log -p` to make sure that each file is appropriately updated.
 
-   `git push && git push origin <tag name>`
+4.  Push the commit and tag:
+
+    git push && git push origin <tag name>
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Please add your schema to this repository under <code>gcn/notices/<i>mission</i>
 The GCN team is happy to iterate with the producers on their schema contents and format.
 Your pipeline will generate JSON files following these schema and send alerts to GCN as described in [New Notice Producers](https://gcn.nasa.gov/docs/producers).
 
-## Version Updates
+## How To Release
 
 New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command. The following steps describe the process for incrementing versions.
 
--- a/README.md
+++ b/README.md
@@ -8,3 +8,7 @@ Please add your schema to this repository under <code>gcn/notices/<i>mission</i>
 
 The GCN team is happy to iterate with the producers on their schema contents and format.
 Your pipeline will generate JSON files following these schema and send alerts to GCN as described in [New Notice Producers](https://gcn.nasa.gov/docs/producers).
+
+## Version Updates
+
+New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command.
--- a/README.md
+++ b/README.md
@@ -16,4 +16,10 @@ Your pipeline will generate JSON files following these schema and send alerts to
    `git clone git@github.com:nasa-gcn/gcn-schema.git `
 
 2. Run `npm version [ major | minor | patch ]`
-   This command will handle the intermediate steps of updating, committing, and pushing the path changes in each file as defined in the `version` and `postversion` npm scripts
+   This command will handle the intermediate steps of updating and committing the path changes in each file as defined in the `version` and `postversion` npm scripts
+
+3. Review the changes with `git log -p` to make sure that each file is appropriately updated.
+
+4. Push the commit and tag:
+
+   `git push && git push origin <tag name>`
--- a/README.md
+++ b/README.md
@@ -11,4 +11,9 @@ Your pipeline will generate JSON files following these schema and send alerts to
 
 ## Version Updates
 
-New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command.
+New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command. The following steps describe the process for incrementing versions.
+
+1. Clone the parent repo (nasa-gcn/gcn-schema). This should not be run in your own forked version of this repo.
+
+2. Run `npm version [ major | minor | patch ]`
+   This command will handle the intermediate steps of updating, committing, and pushing the path changes in each file as defined in the `version` and `postversion` npm scripts
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
     "prepare": "husky install",
     "validate:prettier": "prettier --check .",
     "validate:jsonschema": "node validate.mjs",
-    "validate": "run-s validate:*"
+    "validate": "run-s validate:*",
+    "version": "node update-tags.mjs && git add -u",
+    "postversion": "node update-tags.mjs --reset && git add -u && git commit -m \"Back to development\" && git push && git push --tags"
   },
   "devDependencies": {
     "ajv": "^8.12.0",
--- a/package.json
+++ b/package.json
@@ -1,12 +1,13 @@
 {
   "name": "@nasa-gcn/schema",
+  "version": "0.0.0",
   "scripts": {
     "prepare": "husky install",
     "validate:prettier": "prettier --check .",
     "validate:jsonschema": "node validate.mjs",
     "validate": "run-s validate:*",
     "version": "node update-tags.mjs && git add -u",
-    "postversion": "node update-tags.mjs --reset && git add -u && git commit -m \"Back to development\" && git push && git push --tags"
+    "postversion": "node update-tags.mjs --reset && git add -u && git commit -m \"Back to development\""
   },
   "devDependencies": {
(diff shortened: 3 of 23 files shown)


lpsinger:
> Alternate method as discussed in #101
> 
> @lpsinger , the only concern I have is that this will create the tag/release on the user's fork. Would the user who triggers the update need to be working in a clone of the nasa-gcn repo specifically for this method to work as intended?

Yes. So we should document the release process.

dakota002:
Sounds good, I can update the readme. Will this also require any changes in the repo's settings to allow users to push to main?

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
+import { readFile, writeFile } from 'fs/promises'
+import { exec } from 'child_process'
+import { glob as baseGlob } from 'glob'
+import packageJSON from './package.json' assert { type: 'json' }
+
+const tagPath = `/v${packageJSON.version}/`
+
+async function glob(path) {
+  return await baseGlob(path, {
+    ignore: ['test/**', 'node_modules/**'],
```
lpsinger, line 10: The `test` directory does not exist.

```suggestion
    ignore: ['node_modules/**'],
```

Alternatively, could we get the list of patterns to ignore from `.gitignore`?

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
...
+    exampleFilenames.map(async (path) => {
+      const example = JSON.parse(
+        await readFile(path, {
+          encoding: 'utf-8',
+        })
+      )
+      example['$schema'] = example['$schema'].replace(oldValue, newValue)
+      await writeFile(path, JSON.stringify(example, null, 2))
+    })
+  )
```
lpsinger, line 38: Please refactor to eliminate duplicated code.

Also, I am concerned that the search-and-replace method here could unintentionally overwrite substrings that coincidentally match the Git ref --- for example, `foo.bar.main.Bat`. Admittedly, a contrived corner case, but a corner case nonetheless. Please instead test for whether the value _begins with_ a GitHub URL matching the current branch, and if so, replace _the matching portion at the beginning of the string_.

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
+import { readFile, writeFile } from 'fs/promises'
+import { exec } from 'child_process'
+import { glob as baseGlob } from 'glob'
+import packageJSON from './package.json' assert { type: 'json' }
```
lpsinger, line 4: Is the following legal?

```suggestion
import { version } from './package.json' assert { type: 'json' }
```

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
...
+  )
+}
+
+if (process.argv.some((x) => x == '--reset')) {
+  await loadAndUpdateFiles(tagPath, '/main/')
+} else {
+  await loadAndUpdateFiles('/main/', tagPath)
+}
+
+exec('npx prettier -w .')
```
lpsinger, line 47: Instead of running prettier, can we pass arguments to `JSON.stringify` that ensure that it will match Prettier's settings?

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
...
+          encoding: 'utf-8',
+        })
+      )
+      example['$schema'] = example['$schema'].replace(oldValue, newValue)
+      await writeFile(path, JSON.stringify(example, null, 2))
+    })
+  )
+}
+
+if (process.argv.some((x) => x == '--reset')) {
```
lpsinger, line 41:
```suggestion
if (process.argv.includes('--reset')) {
```

lpsinger (changes requested):

.github/workflows/release.yml
```diff
@@ -0,0 +1,14 @@
...
+      - '*'
+
+jobs:
+  tag-new-versions:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ncipollo/release-action@v1
```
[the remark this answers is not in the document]
dakota002: I'll look into some alternatives

README.md
```diff
@@ -8,3 +8,12 @@ Please add your schema to this repository under <code>gcn/notices/<i>mission</i>
...
 The GCN team is happy to iterate with the producers on their schema contents and format.
 Your pipeline will generate JSON files following these schema and send alerts to GCN as described in [New Notice Producers](https://gcn.nasa.gov/docs/producers).
+
+## Version Updates
+
+New versions are simplified with the [`npm version`](https://docs.npmjs.com/cli/v8/commands/npm-version) command. The following steps describe the process for incrementing versions.
+
+1. Clone the parent repo (nasa-gcn/gcn-schema). This should not be run in your own forked version of this repo.
+
+2. Run `npm version [ major | minor | patch ]`
```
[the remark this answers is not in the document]
dakota002: The push commands are tied into the postversion script which runs automatically with the ncipollo/release-action@v1 action, but if we change this it will need to be updated

update-tags.mjs
dakota002: This threw errors when I tried to implement it, something along the lines of package.json not having a valid export named version

update-tags.mjs
dakota002: Mostly, though it seems like the one specific that is getting in the way is array formatting:

json.stringify will output an array like this:

```
   "enum": [
        "circle",
        "ellipse",
        "rectangle"
      ],
```

where prettier parses it as short enough for one line:

```
"enum": ["circle", "ellipse", "rectangle"],
```

update-tags.mjs
lpsinger: What errors?

update-tags.mjs
lpsinger: OK. Instead of calling it from the CLI, would you please use the API? https://prettier.io/docs/en/api.html

update-tags.mjs
```diff
@@ -0,0 +1,42 @@
...
+async function fileUpdates(fileSet, key, oldValue, newValue) {
+  await Promise.all(
+    fileSet.map(async (fileItem) => {
+      let file = JSON.parse(
+        await readFile(fileItem, {
+          encoding: 'utf-8',
+        })
+      )
+      file[key] = file[key].replace(oldValue, newValue)
+      await writeFile(fileItem, JSON.stringify(file, null, 2).concat('\n'))
```
lpsinger, line 23: No need to adjust the formatting if you do have to run it through prettier.

```suggestion
      await writeFile(fileItem, JSON.stringify(file))
```

update-tags.mjs
```diff
@@ -0,0 +1,42 @@
...
+
+  const exampleFilenames = await glob('**/*.example.json')
+  await fileUpdates(exampleFilenames, '$schema', oldValue, newValue)
+}
+
+if (process.argv.includes('--reset')) {
+  await loadAndUpdateFiles(tagPath, '/main/')
+} else {
+  await loadAndUpdateFiles('/main/', tagPath)
+}
```
lpsinger, line 40: These regexes admit false positive matches. The global search and replace still has a corner case that it doesn't properly handle paths that _happen_ to contain strings that match version tags or branch names but aren't.

lpsinger (changes requested):

update-tags.mjs
dakota002:
```
import { version } from './package.json' assert { type: 'json' }
         ^^^^^^^
SyntaxError: The requested module './package.json' does not provide an export named 'version'
    at ModuleJob._instantiate (node:internal/modules/esm/module_job:123:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:189:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)
    at async loadESM (node:internal/process/esm_loader:91:5)
    at async handleMainPromise (node:internal/modules/run_main:65:12)
```

lpsinger: Oh, but of course! There is no `version` key in your `package.json` file. So add one!

update-tags.mjs
```diff
@@ -0,0 +1,47 @@
+import { readFile, writeFile } from 'fs/promises'
+import { exec } from 'child_process'
```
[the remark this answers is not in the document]
dakota002: Not directly, however it does support the following as a means to convert it to a promise:

```
const util = require('node:util');
const exec = util.promisify(require('node:child_process').exec);

async function lsExample() {
  const { stdout, stderr } = await exec('ls');
  console.log('stdout:', stdout);
  console.error('stderr:', stderr);
}
lsExample();
```

I could implement it in a similar way, but if I make use of the prettier api call this will go away

.github/workflows/release.yml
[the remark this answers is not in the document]
dakota002: Still reviewing alternate options, but [gh-release](https://github.com/marketplace/actions/gh-release) looks like a potential replacement.

.github/workflows/release.yml
```diff
@@ -0,0 +1,14 @@
+name: Create New Release
+on:
+  push:
+    tags:
+      - '*'
```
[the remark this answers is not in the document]
dakota002: I think this is the most generic for allowing any tag, the alternative being '**' https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-event-activity-types

update-tags.mjs
dakota002: Unfortunately that error happens with a `version` key as well:
image

update-tags.mjs
dakota002: True, I have updated the matching (lines 22-25)

package.json
```diff
@@ -4,7 +4,9 @@
     "prepare": "husky install",
     "validate:prettier": "prettier --check .",
     "validate:jsonschema": "node validate.mjs",
-    "validate": "run-s validate:*"
+    "validate": "run-s validate:*",
+    "version": "node update-tags.mjs && git add -u",
+    "postversion": "node update-tags.mjs --reset && git add -u && git commit -m \"Back to development\" && git push && git push --tags"
```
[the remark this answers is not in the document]
dakota002: Do you think it makes sense then to remove the `postversion` command all together?  Or maybe it makes sense to add a command that could be run, like `npm review`, that can display the new version tag and any other possibly relevant information?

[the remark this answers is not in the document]
lpsinger: > Do you think it makes sense then to remove the `postversion` command all together?

No, you still need the "Back to development" commit.

> Or maybe it makes sense to add a command that could be run, like `npm review`, that can display the new version tag and any other possibly relevant information?

No. I think that the instructions should just tell the developer to review at the changes with `git log -p` before pushing them.

update-tags.mjs
lpsinger: How interesting! Well, thanks anyway for trying.

.github/workflows/release.yml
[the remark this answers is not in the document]
dakota002: What are your thoughts on using [action-gh-release](https://github.com/softprops/action-gh-release)? 
Almost 3k stars and fewer files than the other one I was checking out.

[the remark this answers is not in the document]
lpsinger: I like it. Did you audit the code too?

lpsinger:
Remember to re-request a review when you are ready for me to look at this again.

.github/workflows/release.yml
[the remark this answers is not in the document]
dakota002: I did read over the code, it looks like its as straight forward as it gets for this

README.md
```diff
@@ -8,3 +8,18 @@ Please add your schema to this repository under <code>gcn/notices/<i>mission</i>
...
+   `git clone git@github.com:nasa-gcn/gcn-schema.git `
+
+2. Run `npm version [ major | minor | patch ]`
+   This command will handle the intermediate steps of updating and committing the path changes in each file as defined in the `version` and `postversion` npm scripts
+
+3. Review the changes with `git log -p` to make sure that each file is appropriately updated.
+
+4. Push the commit and tag:
+
+   `git push && git push origin <tag name>`
```
lpsinger, line 25:
```suggestion
1.  Clone the parent repo (nasa-gcn/gcn-schema).

        git clone git@github.com:nasa-gcn/gcn-schema.git

2.  Run `npm version [ major | minor | patch ]`

    This command will handle the intermediate steps of updating and committing the path changes in each file as defined in the `version` and `postversion` npm scripts

3.  Review the changes with `git log -p` to make sure that each file is appropriately updated.

4.  Push the commit and tag:

    git push && git push origin <tag name>
```

update-tags.mjs
```diff
@@ -0,0 +1,56 @@
...
+for (const file of files) {
+  const text = await readFile(file, 'utf-8')
+  await writeFile(
+    file,
+    prettier.format(text, {
+      semi: false,
+      singleQuote: true,
+      parser: 'json',
+    })
+  )
```
lpsinger, line 55: Run prettier inside `loadAndUpdateFiles` so that we don't read and write the same files multiple times.

update-tags.mjs
```diff
@@ -0,0 +1,56 @@
...
+  await fileUpdates(schemaFilenames, '$id', oldValue, newValue)
+
+  const exampleFilenames = await glob('**/*.example.json')
+  await fileUpdates(exampleFilenames, '$schema', oldValue, newValue)
+}
+
+if (process.argv.includes('--reset')) {
+  await loadAndUpdateFiles(tagPath, '/main/')
+} else {
+  await loadAndUpdateFiles('/main/', tagPath)
```
lpsinger, line 42:
```suggestion
  await loadAndUpdateFiles(tagPath, 'main')
} else {
  await loadAndUpdateFiles('main', tagPath)
```

update-tags.mjs
```diff
@@ -0,0 +1,56 @@
...
+  await Promise.all(
+    fileSet.map(async (fileItem) => {
+      let file = JSON.parse(
+        await readFile(fileItem, {
+          encoding: 'utf-8',
+        })
+      )
+      file[key] = file[key].replace(
+        `https://gcn.nasa.gov/schema/${oldValue}`,
+        `https://gcn.nasa.gov/schema/${newValue}`
```
lpsinger, line 24:
```suggestion
        `https://gcn.nasa.gov/schema/${oldValue}/`,
        `https://gcn.nasa.gov/schema/${newValue}/`
```

lpsinger (changes requested):

update-tags.mjs
```diff
@@ -0,0 +1,51 @@
...
+      file[key] = file[key].replace(
+        `https://gcn.nasa.gov/schema/${oldValue}/`,
+        `https://gcn.nasa.gov/schema/${newValue}/`
+      )
+
+      await writeFile(
+        fileItem,
+        prettier.format(JSON.stringify(file), {
+          semi: false,
+          singleQuote: true,
```
lpsinger, line 31: I doubt that either the `semi` or `singleQuote` options have any effect on JSON documents.

lpsinger (changes requested):

update-tags.mjs
dakota002: Yea, just removed them

lpsinger (approved):

PR #108, opened 2023-09-25
Include null as option for p-values

blaufuss:
Make it clear that p-values reported (overall search or per-event one…s) can be null.  This happens when a search is not run for selection reasons, or an individual event is not significant in one search or the other.

### DIFF
--- a/gcn/notices/icecube/LvkNuTrackSearch.example.json
+++ b/gcn/notices/icecube/LvkNuTrackSearch.example.json
@@ -23,7 +23,7 @@
         "systematic_included": false
       },
       "event_pval_generic": 0.0191,
-      "event_pval_bayesian": 0.0549
+      "event_pval_bayesian": null
     },
     {
       "event_dt": 222.46,
--- a/gcn/notices/icecube/LvkNuTrackSearch.schema.json
+++ b/gcn/notices/icecube/LvkNuTrackSearch.schema.json
@@ -23,11 +23,11 @@
   "properties": {
     "pval_generic": {
       "description": "P-value from generic transient search; consistency with background expectations [0-1]",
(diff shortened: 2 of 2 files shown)


lpsinger:
Is it OK to just omit the field when it is unknown, rather than set it to null? If so, then no changes are needed.

blaufuss:
Our preference is to use these nulls to indicate things that aren't reported, but if you would prefer not to use nulls, let us know and we will adapt.

Vidushi-GitHub:
For IceCube specific-fields in schema, it's producer's preference.

lpsinger (approved):

blaufuss:
Thanks all, is it possible to get a new stable tag for the schema issued?  @Vidushi-GitHub or @lpsinger ?

lpsinger:
Sure. @dakota002, are there any other changes to include in the next release?

Vidushi-GitHub:
Two changes: (a) Only number to union for core/follow-up schema, not breaking for existing IceCube/Guano and also, works well for new IceCube. (b) Ice-Cube LVK schema

lpsinger:
I think that we ought to do a major version bump.

Vidushi-GitHub:
> I think that we ought to do a major version bump.

2.0.0 or 2.0.1? as, core and patch change

lpsinger:
No, _just_ a major version change, 2.0.0.

Vidushi-GitHub:
Stable tag v2.0.0 is there @blaufuss. @dakota002 @lpsinger

blaufuss:
> Stable tag v2.0.0 is there @blaufuss. @dakota002 @lpsinger

Thanks, much appreciated.

PR #113, opened 2023-10-18
Guano notice type in bat directory

Vidushi-GitHub:
@Tohuvavohu please check this. As Guano is specific pipeline/notice type of bat, makes sense in bat directory. @jracusin


Tohuvavohu:
LGTM

lpsinger (approved):

PR #119, opened 2023-12-21
new schema for einstein probe

modot:
I added the first notice schema for the Einstein Probe Satellite. An example is included.

Please review the changes and provide any feedback or suggestions. Thank you for your time and consideration.


dakota002 (approved):

Vidushi-GitHub (commented):
Looks good! 

I suggest using the Statistics schema from the core instead of defining specific properties like src_significance. Statistics schema in core includes rate_snr and rate_energy_range. Also, please consider using energy range as a separate property instead of including it in the description.

Would you like to add detector information in schema?

gcn/notices/ep/wxt/alert.schema.example.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/ep/wxt/alert.schema.json",
+  "mission": "EP",
+  "instrument": "WXT",
+  "trigger_time": "2024-03-01T21:46:05.13Z",
+  "ra": 120,
+  "dec": 40,
+  "ra_uncertainty": [0.02],
```
lpsinger, line 8: @Vidushi-GitHub, could we please permit this to be a single number instead of a list?

Vidushi-GitHub: It's kept generic purposely, to accommodate elliptical, rectangle uncertainties. Ofcourse, we can keep it flexible (number/string) using Object type.

lpsinger: Please open a PR to support a union of either a list or a single number.

modot:
> Looks good!
> 
> I suggest using the Statistics schema from the core instead of defining specific properties like src_significance. Statistics schema in core includes rate_snr and rate_energy_range. Also, please consider using energy range as a separate property instead of including it in the description.

Thank you for your suggestions! In the new commit, I have made the changes by replacing src_significance with img_snr and utilizing the energy range property.
> 
> Would you like to add detector information in schema?

Do you mean to use the detector status property? I've confirmed with our team that there's no need to include this property.

Vidushi-GitHub:
Hi @modot, we have made modification in [Localization schema](https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/core/Localization.schema.json), please replace the property "ra_uncertainty" to "ra_dec_error". We will tag the new version soon.

gcn/notices/ep/wxt/alert.schema.example.json
```diff
@@ -0,0 +1,12 @@
...
```
lpsinger, line 8:
```suggestion
  "ra_dec_error": [0.02],
```

lpsinger (changes requested):

modot:
> @modot, we've made a small change to how we represent position errors --- see above, [#119 (comment)](https://github.com/nasa-gcn/gcn-schema/pull/119#discussion_r1507902309). Would you please acknowledge that this is OK?

This is OK, I've made the suggested change.

gcn/notices/ep/wxt/alert.schema.example.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/ep/wxt/alert.schema.json",
+  "mission": "EP",
```
lpsinger, line 3: You can safely remove this, because it's implied in the Kafka topic name.

```suggestion
```

gcn/notices/ep/wxt/alert.schema.example.json
lpsinger, line 1: Please rename this directory from `ep` to `einstein_probe` to match the Kafka topic name.

lpsinger (changes requested):

gcn/notices/ep/wxt/alert.schema.example.json
modot: Ok, will do.

lpsinger (approved):

PR #120, opened 2023-12-29
Support a list or a single number in localization

Vidushi-GitHub:


lpsinger (changes requested):
Can we have a single description for each field, like this?

```json
"ra_uncertainty": {
  "description": "fill me in",
  "oneOf": [
    {"type": "number"},
    {"type" "array", "maxItems": 2, "items": {"type": "number"}}
  ]
}
```
```

File: gcn/notices/core/Localization.schema.json [hunk 1]
```diff
@@ -19,14 +19,18 @@
       "default": "circle"
     },
     "ra_uncertainty": {
-      "type": "array",
-      "maxItems": 2,
-      "description": "Uncertainty in RA [deg] with optional asymmetric errors, measured from the RA in the schema; respective ex: [r], [r1], [l1-, l1+]. For ellipse, ra axis is r1 or semi-major axis."
+      "description": "Uncertainty in RA [deg], measured w.r.t. RA. A single number for circle, and ellipse, RA axis alligns with semi-major axis. For rectangle, use optional aymmetric errors array, [l1-, l2+].",
```
lpsinger, line 22: What does "Uncertainty in RA measured with respect to RA" mean?

Vidushi-GitHub: So there have been confusion, about it's understanding. It's uncertainty error measured from RA centre like 2 (-+0.2) or, shape coordinates 2 (1.8, 2.2), may be you can help with better wordings.

lpsinger: In that case, the name of this field is wrong. It is the uncertainty in the semimajor axis.

Vidushi-GitHub: But but if it's rectangle?

Vidushi-GitHub, line 22:
```suggestion
      "description": "Uncertainty in RA [deg]. A single number for circle, and ellipse, RA axis alligns with semi-major axis. For rectangle, use optional aymmetric errors array, [l1-, l2+].",
```

gcn/notices/core/Localization.schema.json
```diff
@@ -19,14 +19,18 @@
...
+      "oneOf": [
+        { "type": "number" },
+        { "type": "array", "maxItems": 2, "items": { "type": "number" } }
+      ]
     },
     "dec_uncertainty": {
-      "type": "array",
-      "maxItems": 2,
-      "description": "Uncertainty in Dec [deg] with optional asymmetric errors, measured from the Dec in the schema; respective ex: [r2], [l2-, l2+]. Not reported if circular and Dec axis is r2 or semi-minor axis for ellipse."
+      "description": "Uncertainty in Dec [deg], measured w.r.t. Dec, not reported for circle. A single number for ellipse, Dec axis alligns with semi-minor axis. For rectangle, use optional aymmetric errors array, [b1-, b2+].",
```
Vidushi-GitHub, line 29:
```suggestion
      "description": "Uncertainty in Dec [deg], not reported for circle. A single number for ellipse, Dec axis alligns with semi-minor axis. For rectangle, use optional aymmetric errors array, [b1-, b2+].",
```

lpsinger:
I spoke with @jracusin about this. I think that the positional uncertainty fields are poorly conceived and named and need a redesign. What we call `ra_uncertainty` and `dec_uncertainty` are really the semi major and semi minor axes of an ellipse containing a given amount of probability (which I believe we left unspecified, another problem).

Please study how these fields are named and encoded in other prominent software applications in astronomy, and propose names and descriptions that are as consistent as possible with prior work:
- VOEvent
- SExtractor, Scamp, Swarm
- IRAF

Vidushi-GitHub:
Yes, had a discussion with @jracusin. She agreed to correct the description. Vera Rubin uses uncertainty word. Elliptical uncertainty ex: https://gcn.nasa.gov/circulars/35535. We can't shift the center, to make errors symmetric, as mean can be different than negative/positive bars, ex bayesian posterior cases.

lpsinger:
This change is not an improvement over the current situation. There should be two positional uncertainty fields for uncertainty in orthogonal directions and one position angle field. All three fields should be scalar.

You give two reasons for having array-valued positional uncertainties, but I find neither convincing:
- *Representation of IPN localizations.* IPN localizations are either annuli (in the case of two satellites), parallelograms (three spacecraft), or irregular polygons (four or more spacecraft). They are not rectangles, and they cannot be described using the parametrization here. We should _not_ define a vocabulary to define arbitrarily complex region geometry.
- *Representation of localizations that are not centrally concentrated.* Without a quantitative definition of the moments of the distribution in terms of the uncertainty parameters, this is too imprecise to be useful.

I insist that the only localization parametrization that we should support (other than a freeform HEALPix map) is a bounding ellipse. Based on prior astronomy software that reports position uncertainty on points, I strongly recommend giving scalar three parameters: the length of the semi major axis, the length of the semi minor axis, and position angle of an ellipse containing N% probability. If the semi minor axis is omitted, then it is assumed to equal the semi major axis. If the position angle is omitted, then it is assumed to be zero.

Vidushi-GitHub:
What about following? I am still trying more options

"uncertainty_shape": {
      "enum": ["circle", "ellipse"],
      "description": "Shape of the uncertainty region for RA and Dec errors. Choose between 'circle' and 'ellipse'.",
      "default": "circle"
    },
    "ra_uncertainty": {
      "description": "Uncertainty in RA [degrees]. Provide the radius for 'circle'. For 'ellipse', provide semi-major axes, where RA is assumed to aligned along the semi-major axis.",
      "oneOf": [
        { "type": "number" },
        { "type": "array", "maxItems": 2, "items": { "type": "number" } }
      ]
    },
    "dec_uncertainty": {
      "description": "Uncertainty in Dec [degrees]. Do not report for 'circle'. For 'ellipse', provide semi-minor axes, such that semi-minor axis representing the Dec value.",
      "oneOf": [
        { "type": "number" },
        { "type": "array", "maxItems": 2, "items": { "type": "number" } }
      ]
    },
    "position_angle": {
      "type": "number",
      "description": "Anti-clockwise position angle from the RA axis [degrees]. Used if uncertainty_shape is 'ellipse' and tilted with respect to the standard (RA, Dec) coordinate system.",
      "default": 0
    },

gcn/notices/core/Localization.schema.json
```diff
@@ -13,24 +13,17 @@
       "type": "number",
       "description": "ICRS declination [deg], utilizes the J2000 epoch and an equatorial coordinate system"
     },
-    "uncertainty_shape": {
-      "enum": ["circle", "ellipse", "rectangle"],
-      "description": "In what manner are the errors in RA and Dec defined - as circular, elliptical, or rectangular shapes",
-      "default": "circle"
+    "position_a_error": {
+      "description": "Radii of the semi-major axis [deg]. ",
```
jracusin, line 17:
```suggestion
      "description": "Length of the semi-major axis [deg]. ",
```

gcn/notices/core/Localization.schema.json
```diff
@@ -13,24 +13,17 @@
...
+      "type": "number"
     },
-    "ra_uncertainty": {
-      "type": "array",
-      "maxItems": 2,
-      "description": "Uncertainty in RA [deg] with optional asymmetric errors, measured from the RA in the schema; respective ex: [r], [r1], [l1-, l1+]. For ellipse, ra axis is r1 or semi-major axis."
+    "position_b_error": {
+      "description": "Radii of the semi-minor axis [deg] of the ellipse. Do not report for circle.",
```
jracusin, line 21:
```suggestion
      "description": "Length of the semi-minor axis [deg] of the ellipse. Do not report for circle.",
```

Vidushi-GitHub:
May look redundant, but just for simplicity, I have tried to keep Josh feedback as it is. I feel may be in future community start shifting to proposal 2.

gcn/notices/swift/bat/guano.loc_arc_min.example.json
```diff
@@ -10,7 +10,7 @@
   "trigger_time": "2022-12-31T21:46:05.13Z",
   "ra": 336.26,
   "dec": 25.139,
-  "ra_uncertainty": [0.05],
+  "error": { "radius": 0.5 },
```
jracusin, line 13:
```suggestion
  "error": { "radius": 0.05 },
```

gcn/notices/core/Localization.schema.json
```diff
@@ -13,25 +13,41 @@
...
+          },
+          "required": [
+            "position_error_a",
+            "position_error_b",
+            "position_error_pa"
+          ]
+        }
+      ]
     },
     "containment_probability": {
```
jracusin, line 52: If we go with this approach, containment_probability should also be part of error.

gcn/notices/core/Localization.schema.json
```diff
@@ -13,25 +13,41 @@
...
-    "dec_uncertainty": {
-      "type": "array",
-      "maxItems": 2,
-      "description": "Uncertainty in Dec [deg] with optional asymmetric errors, measured from the Dec in the schema; respective ex: [r2], [l2-, l2+]. Not reported if circular and Dec axis is r2 or semi-minor axis for ellipse."
-    },
-    "position_angle": {
-      "type": "number",
-      "description": "Anti-clockwise position angle from RA axis [deg]. Used if uncertainty_shape ellipse or rectangle is tilted w.r.t. standard (RA, Dec) coordinate.",
-      "default": 0
+    "error": {
```
jracusin, line 16: 'error' is a bit too generic, there can be error in many quantities.  We could call it position_error.  This nesting seems to break the flattening of the schema, and be a bit clunky.

lpsinger (changes requested):
@Vidushi-GitHub, I asked you to make `position_error` a union of a single number or an array, with the elements of that array being a, b, and pa. This is not an array; it's a composite data structure. None of our other core schema contain composite data types. Let's stick to that convention.

lpsinger (changes requested):

PR #141, opened 2024-04-24
heasarc - adding archive noitices

krutkow:
Added the schema to create notices for when data hit the heasarc archive.
This is all being done with ixpe for now, and all testing will be done with ixpe. I included the nustar dir just to see how this can scale in the directory structure.


lpsinger (commented):
These three schema seem like they are almost identical. Could we have just a single schema for these various notice types?

README.md
```diff
@@ -4,7 +4,7 @@ This repository contains the schemas for crafting new GCN notices. It also inclu
 
 The alert format for new Notices over GCN Kafka is JavaScript Object Notation (JSON), a widely adopted Internet data format. We have designed a set of core schema which serve as the building blocks for new GCN Notices. Instrument-specific schema can also be created, but we request that you utilize the core schema as much as possible.
 
-Please add your schema to this repository under <code>gcn/notices/<i>mission</i>/</code> and submit a pull request for the GCN Team to review. The GCN team is happy to iterate with the producers on their schema contents and format. Your pipeline will generate JSON files following these schema and send alerts to GCN as described in [New Notice Producers](https://gcn.nasa.gov/docs/producers).
+Please add your schema to this repository under <code>gcn/notices/<i>mission</i>/</code> and submit a pull request for the GCN Team to review. The GCN team is happy to iterate with the producers on their schema contents and format. Your pipeline will generate JSON files following these schema and send alerts to GCN as described in [New Notice Producers](https://gcn.nasa.gov/docs/notices/producers).
```
lpsinger, line 7: Nice catch. Would you please open a separate PR for that?

gcn/notices/heasarc/archive_alert.schema.json
```diff
@@ -0,0 +1,37 @@
...
+      "description": "The observation ID of the sequence"
+    },
+    "target_name": {
+      "type": "string",
+      "description": "Name of the target object for the observation"
+    },
+    "observation_datetime": {
+      "type": "string",
+      "description": "Start time of observation"
+    },
```
lpsinger, line 31: We probably have core schema fields that you can use for these.

lpsinger:
@dakota002, would you please help us figure out why the validation test is not passing?

krutkow:
I just figured it out - I had the unevaluatedProperties set to false, so validation failed when it saw my defined properties. I removed that and it worked.

dakota002:
> I just figured it out - I had the unevaluatedProperties set to false, so validation failed when it saw my defined properties. I removed that and it worked.

Ah I see, so this was failing because there is a $schema field in your example file. This is an error in our documentation that we should clarify. `'unevaluatedProperties': false` is saying that there should not be any additional properties than the ones defined in the schema. To include this correctly, we should have specified that you also just need to add `"$schema": true,` to your properties, otherwise it is parsed as an unevaluated property. 

I will make an update for our docs, good catch!

gcn/notices/heasarc/archive_alert.schema.json
```diff
@@ -0,0 +1,39 @@
...
+      "type": "string",
+      "description": "The observation identifier of the sequence. Also called the observation id, sequence number, and seqnum, it is a series of numerical digits, each signifying different aspects for the observation, such as AO year, proposal number, and target number. Each mission has their own specification for how the obsid is defined. Every observation by an observatory is identified by the obsid. The data set identified by the obsid contains the level one and level 2 data created by that mission's processing pipeline. [Fields to align with IVOA ObsCore]"
+    },
+    "target_name": {
+      "type": "string",
+      "description": "Name of the target object for the observation [IVOA ObsCore]"
+    },
+    "mission_page": {
+      "type": "string",
+      "description": "Webpage of the creator mission at HEASARC"
```
lpsinger, line 32:
```suggestion
      "description": "Web page of the creator mission at HEASARC"
```

lpsinger (changes requested):

lpsinger (approved):

PR #147, opened 2024-05-31
Naming fixes

dakota002:
Naming fixes for example files

PR #149, opened 2024-05-31
guano schema: snake_case

Vidushi-GitHub:
# Description
For consistency, Guano schema files names are modified to snake_case format.
These are just examples, should n't break anything.

Somehow id field is in correct format: "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/swift/bat/guano.schema.json"
Notice is already in lower case: "gcn.notices.swift.bat.guano"

Related PR:
https://github.com/nasa-gcn/gcn-schema/pull/147

P.S. See if we wanna confirm this with someone from Swift team?


dakota002 (changes requested):
@Vidushi-GitHub if you change the schema name to be lowercase ('guano') you need to update the `$schema` properties so that it is still valid, which means this is also technically a breaking change.

Vidushi-GitHub:
> @Vidushi-GitHub if you change the schema name to be lowercase ('guano') you need to update the `$schema` properties so that it is still valid, which means this is also technically a breaking change.

It dint seem breaking for validation.

lpsinger:
> > @Vidushi-GitHub if you change the schema name to be lowercase ('guano') you need to update the `$schema` properties so that it is still valid, which means this is also technically a breaking change.
> 
> It dint seem breaking for validation.

That's a bug in our validation, then. URLs are generally case-sensitive.

PR #155, opened 2024-06-20
Add EP event name

Vidushi-GitHub:
# Description
Following the Zendesk ticket #210 (https://nasa-gcn.zendesk.com/agent/tickets/210): 
The same event is being notified through both Notices and Circulars. However, they don't contain a common unique ID and have slightly different times and localization (as Circulars are more refined). 

It is necessary to include a property such as the event name or trigger ID as a unique ID, which should be the same across both Notices and Circulars. 
One potential option is the event name, which is currently reported in EP Circulars, and request to add it to the Notices pipeline. 

@modot, please check what's more convenient at your end, either include the trigger ID or event name in the EP Notices.


Vidushi-GitHub:
Purpose solved in https://github.com/nasa-gcn/gcn-schema/pull/160

PR #161, opened 2024-07-11
GECAM alert schema and example

athish-thiru:
The commit has two files: a gecam schema json and an example json


gcn/notices/classic/gecam/alert.example.json
lpsinger, line 1: Please move this file to gcn/notices/gecam/alert.example.json.

gcn/notices/classic/gecam/alert.schema.json
lpsinger, line 1: Please move this file to gcn/notices/gecam/alert.schema.json.

lpsinger (changes requested):

athish-thiru:
I have moved the files to the new location, updated $id and $schema, and removed the classic directory.

lpsinger:
> I have moved the files to the new location, updated $id and $schema, and removed the classic directory.

Oh, oops, I misunderstood. I understand now that these are converted from GCN Classic. I am not sure about where the schema should live.

PR #162, opened 2024-07-17
Rename Orbit.schema.json to match $id

lpsinger:


Vidushi-GitHub:
I agree with the changes, $id should be consistent with the filename. 

According to our unified schema document, we decided to keep mission names in snake_case. 
Do you want to keep the capital letter?

tylerbarna (approved):

lpsinger:
> According to our unified schema document, we decided to keep mission names in snake_case.
> Do you want to keep the capital letter?

Not in this PR. Any changes to the schema filenames would need to be approved by the corresponding missions' points of contact.

PR #165, opened 2024-07-18
Make ajv-formats a dev-only dependency

lpsinger:
It is only used for validation.


dakota002 (approved):

dakota002:
The changes look good, just needs a rebase

PR #176, opened 2024-07-22
Get latest tag

dakota002:
`git describe --exact-match --tags` returns the error:
`fatal: no tag exactly matches '5c6a98ef76738659b1a1dcf5666744ba466687ee'`

This should resolve that

Edit:
`git describe --abbrev=0` also returns the latest tag, effectively the same as git describe --tags --abbrev=0


Vidushi-GitHub (approved):

lpsinger:
This broke validation on main.

PR #194, opened 2024-09-06
Schema and example drafts for the KM3NeT alert programm

VincentLuminy:
# Description
First draft of KM3NeT alert schema and example. Not yet in production (no publicly available alerts) and subject to evolution.

# Testing

```
import json
from jsonschema import validate, ValidationError

# Load JSON schema from file
with open('./gcn/notices/km3net/test/medal_ranking_alert.schema.json', 'r') as file:
    schema = json.load(file)

# Load JSON example from file
with open('./gcn/notices/km3net/test/medal_ranking_alert.example.json', 'r') as file:
    json_data = json.load(file)

# Validate that example match schema
try:
    validate(instance=json_data, schema=schema)
    print("JSON valide")
except ValidationError as e:
    print(f"JSON invalide : {e.message}")
```

### DIFF
--- a/gcn/notices/km3net/test/medal_ranking_alert.example.json
+++ b/gcn/notices/km3net/test/medal_ranking_alert.example.json
@@ -7,7 +7,7 @@
   "alert_tense": "test",
   "alert_type": "initial",
   "alert_datetime": "2024-09-01T12:01:00.00Z",
-  "analysis_pipeline": "exceptionnal_evt_arca",
+  "analysis_pipeline": "exceptional_evt_arca",
   "description": "KM3NeT online analysis, bronze candidate neutrino observation.",
   "event_name": ["KM240901A"],
   "trigger_time": "2024-09-01T12:00:00.00Z",
--- a/gcn/notices/km3net/test/medal_ranking_alert.example.json
+++ b/gcn/notices/km3net/test/medal_ranking_alert.example.json
@@ -1,0 +1,32 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/km3net/test/medal_ranking_alert.schema.json",
+  "mission": "KM3NeT",
+  "instrument": "ARCA028",
+  "messenger": "Neutrino",
+  "packet_type": 999,
+  "alert_tense": "test",
+  "alert_type": "initial",
+  "alert_datetime": "2024-09-01T12:01:00.00Z",
+  "analysis_pipeline": "exceptionnal_evt_arca",
+  "description": "KM3NeT online analysis, bronze candidate neutrino observation.",
+  "event_name": ["KM240901A"],
+  "trigger_time": "2024-09-01T12:00:00.00Z",
+  "ra": 13.82,
+  "dec": 19.01,
+  "ra_dec_error": 0.9,
+  "healpix_url": "https://www.km3net.org/about-km3net/open-access/",
+  "far": 8.029e-8,
+  "additional_info": "Track only / Track+Shower analysis. Up-going / All-sky selection. Analysis pipeline event selection tuned to select X event per month in average.",
+  "triggering_evts": [
+    {
+      "trigger_time": "2024-09-01T12:00:00.00Z",
+      "ra": 13.82,
+      "dec": 19.01,
+      "ra_dec_error": 0.9,
+      "cos_zenith": -0.95,
+      "prob_evt_topology": "(Starting) Track / Shower",
+      "n_triggered_pmt": 586
+    }
+  ],
+  "known_sources": "No search / No match / [List]"
+}
--- a/gcn/notices/km3net/test/medal_ranking_alert.schema.json
+++ b/gcn/notices/km3net/test/medal_ranking_alert.schema.json
@@ -81,15 +81,15 @@
   "$defs": {
     "prob_evt_topology": {
       "type": "string",
-      "description": "track, shower, starting event."
+      "description": "Most probable event topology, could be track ('starting' if the origin is inside the detector) or shower. Track is associated to a muon flavor, shower can be the 3 flavors."
     },
     "cos_zenith": {
       "type": "number",
-      "description": "Value of the cosine of the angle between the event direction and the vertical -1 correspond to up-going event while 1 is down-going."
+      "description": "Value of the cosine of the angle between the event direction and the vertical. -1 correspond to up-going event while 1 is down-going."
     },
     "n_triggered_pmt": {
       "type": "number",
-      "description": "Number of PMT triggered by the event (energy proxy)."
+      "description": "Number of PMT triggered by the event (i.e energy proxy)."
     },
     "space_time_coord": {
       "type": "object",
--- a/gcn/notices/km3net/test/medal_ranking_alert.schema.json
+++ b/gcn/notices/km3net/test/medal_ranking_alert.schema.json
@@ -1,0 +1,142 @@
+{
+  "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/km3net/test/medal_ranking_alert.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "unevaluatedProperties": false,
+  "title": "KM3NeT/Gold-Silver-Bronze Alerts",
+  "description": "Candidate neutrino observation reported from the KM3NeT online analysis pipeline.",
+  "allOf": [
+    {
+      "$ref": "../../core/Reporter.schema.json",
+      "description": "Instrument (ARCAxxx/ORCAxxx) issuing the alert. Mission=KM3NeT; Messenger=Neutrino."
+    },
+    {
+      "$ref": "../../core/Alert.schema.json",
+      "description": "Tense role (test/current observation), type (initial/update/retraction) and datetime of the alert generation"
+    },
+    {
+      "$ref": "../../core/Event.schema.json",
+      "description": "Event ID (eg 1A. Nb increment for each new alert, letter increment for update/retraction of the nb). Event_Name (KM3-yymmddL) if appropriate. "
+    },
+    {
+      "$ref": "#/$defs/space_time_coord",
+      "description": "Most probable Localization (Time, Position, Error and Skymap if available)."
+    },
+    {
+      "$ref": "../../core/Statistics.schema.json",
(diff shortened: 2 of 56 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@nasa-gcn/schema",
   "description": "Unified JSON Schema for multi-messenger and high-energy astronomy alerts",
-  "version": "4.1.0",
+  "version": "4.2.0",
```
lpsinger, line 4:
```suggestion
  "version": "4.1.0",
```

Don't bump the version. We'll do that when we do a release.

lpsinger (changes requested):

package.json
VincentLuminy: Sorry, I think it results from one of the readme commands (maybe when I did npm `version minor`). I will try to correct it.

VincentLuminy: Hi, I have applied your suggestion and commit it. Is it enough to solve the problem or do you need me to do something else?
In my commit, this problem occurs twice, one in package.json and one in package-lock.json. If the problem is not solved, I saw that I can delete the files from my commit, would that solve the problem?

gcn/notices/km3net/test/medal_ranking_alert.schema.json
```diff
@@ -0,0 +1,142 @@
...
+      "items": {
+        "anyOf": [
+          { "$ref": "#/$defs/knownsrc" },
+          { "$ref": "#/$defs/mmcoincevent" }
+        ]
+      },
+      "description": "No search / No match / Array of known astrophysical sources present in the KM3NeT event RoI (not exhaustive) OR reference and position for a coincident time-variable source."
+    }
+  },
+  "required": [
```
Vidushi-GitHub, line 61: If you with you can skip "required" option, it will enforce all fields must be there.

gcn/notices/km3net/test/medal_ranking_alert.schema.json
```diff
@@ -0,0 +1,142 @@
...
+    "trigger_time",
+    "ra",
+    "dec",
+    "ra_dec_error",
+    "far",
+    "triggering_evts",
+    "known_sources"
+  ],
+  "$defs": {
+    "prob_evt_topology": {
```
Vidushi-GitHub, line 82: Description should be clear for users

gcn/notices/km3net/test/medal_ranking_alert.example.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/km3net/test/medal_ranking_alert.schema.json",
+  "mission": "KM3NeT",
+  "instrument": "ARCA028",
+  "messenger": "Neutrino",
+  "packet_type": 999,
+  "alert_tense": "test",
+  "alert_type": "initial",
+  "alert_datetime": "2024-09-01T12:01:00.00Z",
+  "analysis_pipeline": "exceptionnal_evt_arca",
```
Vidushi-GitHub, line 10:
```suggestion
  "analysis_pipeline": "exceptional_evt_arca",
```

gcn/notices/km3net/test/medal_ranking_alert.example.json
```diff
@@ -0,0 +1,32 @@
...
+  "description": "KM3NeT online analysis, bronze candidate neutrino observation.",
+  "event_name": ["KM240901A"],
+  "trigger_time": "2024-09-01T12:00:00.00Z",
+  "ra": 13.82,
+  "dec": 19.01,
+  "ra_dec_error": 0.9,
+  "healpix_url": "https://www.km3net.org/about-km3net/open-access/",
+  "far": 8.029e-8,
+  "additional_info": "Track only / Track+Shower analysis. Up-going / All-sky selection. Analysis pipeline event selection tuned to select X event per month in average.",
+  "triggering_evts": [
```
Vidushi-GitHub, line 20: How this ra,dec is different from the event above?

VincentLuminy: In this case (where the alert is produced by only one event) there is no difference. 
But we have analysis pipeline where several events (coming from the same direction in a short time window, but individually below our threshold to send alert) would be identified as a signal of interest and lead to alert creation. 
In that case, the coordinates in "triggering_evts" are the coordinates of our individual events, while the "ra", "dec" and "ra_dec_error" of the alert body correspond to the most probable direction. 

We wanted to have an alert message as generic as possible, this is why we decided to duplicate the information, but I am open to suggestions if you have.

Vidushi-GitHub: Ok thanks.
Localization (ra,dec) is used at two places. 

(---)"ra": 13.82,
  "dec": 19.01,
  "ra_dec_error": 0.9,
  "healpix_url": "https://www.km3net.org/about-km3net/open-access/",
  "far": 8.029e-8,
  "additional_info": "Track only / Track+Shower analysis. Up-going / All-sky selection. Analysis pipeline event selection tuned to select X event per month in average.",
  "triggering_evts": [
{
      "trigger_time": "2024-09-01T12:00:00.00Z",
(---)      "ra": 13.82,
      "dec": 19.01,

I would suggest, either put all localizations in trigger_evts (ra, dec) list, Or create two examples files for one schema. One for alert, another for several events analysis pipeline.
It's creating confusion, example, whether above one is follow-up event.

gcn/notices/km3net/test/medal_ranking_alert.schema.json
Vidushi-GitHub: If your pipeline will produce all the fields for each notices, then fine.
One can read at schema browser (https://gcn.nasa.gov/docs/schema/v4.1.0/gcn/notices), for what to expect, even we will soon create and announce mission page soon after your feedback.
eg: https://gcn.nasa.gov/missions/einstein-probe
It will put constraint on your pipeline, it's upto you.

Vidushi-GitHub:
Hi Vincent, in continuation of your Notices type, your plan (email reference) is "Regarding the topic name, we are planning to send three types of notices: Gold, Silver, and Bronze. However, we are using a single schema called "medal_ranking_alert." I hope it won't be too much of a problem." 
Sounds good for keeping one topic. However in schema would you like to add medal_rank property with enum [bronze, gold, silver] ?

gcn/notices/km3net/test/medal_ranking_alert.example.json
```diff
@@ -0,0 +1,32 @@
...
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/km3net/test/medal_ranking_alert.schema.json",
+  "mission": "KM3NeT",
+  "instrument": "ARCA028",
+  "messenger": "Neutrino",
+  "packet_type": 999,
+  "alert_tense": "test",
+  "alert_type": "initial",
+  "alert_datetime": "2024-09-01T12:01:00.00Z",
+  "analysis_pipeline": "exceptional_evt_arca",
+  "description": "KM3NeT online analysis, bronze candidate neutrino observation.",
```
Vidushi-GitHub, line 11: Make it specific field, like medal_rank, instead of description. It is useful property, should be machine-readable.

gcn/notices/km3net/test/medal_ranking_alert.schema.json
```diff
@@ -0,0 +1,142 @@
...
+    {
+      "$ref": "../../core/AdditionalInfo.schema.json",
+      "description": "Description of the pipeline used to produce the alert."
+    }
+  ],
+  "properties": {
+    "$schema": true,
+    "packet_type": {
+      "type": "number",
+      "description": "packet_type provided by GCN as for the old GCN format, associatied to notice_type/topic Gold=, Silver=, Bronze=."
```
Vidushi-GitHub, line 38: Is this associated with your pipeline? 
Or, GCN Classic? As we will not rely on the old system in the future.

@jracusin would be have packet type with new system?
KM3NeT had following concern over email:"Additionally, my collaboration has requested that we include a packet_type in our alerts. Would it be possible to provide us with three distinct packet_type values (one for each of the Gold, Silver, and Bronze alerts), similar to how it was done with the classical VOEvent messages?"

VincentLuminy: As you stated, we would like to be able to give a packet_type in the notice, like it was done in GCN Classic. The number here are random for example, but will be updated with the accurate value if we have one.

gcn/notices/km3net/test/medal_ranking_alert.example.json
VincentLuminy: Indeed it could be confusing. I will create a second example file with several events in the alert. We are also discussing to rename "triggering_evts" into something more explicit ("hits_information" or "internal_triggers_information" or something else). I will commit the change as soon as we reach an internal agreement. 
The descriptions in the schema file will be updated as well (both for the alert coordinates and the triggering_evts variable).

gcn/notices/km3net/test/medal_ranking_alert.schema.json
VincentLuminy: Hi @Vidushi-GitHub @jracusin would it be possible to have a packet_type even if our notices are not part of GCN classic ? 
I know it is not needed from a technical point of view with Kafka, but we want to provide this information for compatibility with pipelines already existing in experiments susceptible to receive our alerts.

jracusin: @VincentLuminy We'd rather not reference back to GCN Classic notice formats.  We can definitely find a way to list your classification of "Gold, Silver, Bronze", but "packet_type" isn't the most descriptive or intuitive keyword.  

You could use the [classification in the statistics core schema](https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/core/Statistics.schema.json), though that is designed to have a probabilistic classification for each choice as a dictionary.  

You could use "additionalInfo" like [IceCube](https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/icecube/test/gold_bronze_track_alerts.example.json). 

Or you could create a custom schema keyword like you have done, but maybe name it something more intuitive like "alert_criteria" or "alert_credibility".

VincentLuminy:
The format in the test folder is depreciated. A new directory architecture will be updated in a future PR.

PR #199, opened 2024-09-20
Fermi

shb46:
Fermi GBM Schema for a single schema, "ALERT" - Initial Version


gcn/notices/fermi/gbm/FermiGBMAlert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    }
+  ],
+  "properties": {
+    "notice_type": {
+      "type": "string",
+      "description": "Notice title"
+    },
+    "hitl": {
+      "type": "boolean",
+      "description": "Human in the loop"
```
Vidushi-GitHub, line 40: Could you please describe properly, not understandable for user?

Vidushi-GitHub:
We have planned a mission page, where you can keep the content present at readme file.

Vidushi-GitHub:
Hi @shb46, sorry for the delayed response due to travel and vacation. Most of the stuff looks fine. 
My main confusion about schema is why strict schema there and why you are re-writing core properties, instead of ref?

shb46:
Hi @Vidushi-GitHub,

Hope everything is well!  Here are my replies:

> We have planned a mission page, where you can keep the content present at readme file.

* Can we have an instrument page as well?  the lat and the GBM instruments are run by separate teams.
* Can notice type have its own page?

> Hi @shb46, sorry for the delayed response due to travel and vacation. Most of the stuff looks fine. My main confusion about schema is why strict schema there and why you are re-writing core properties, instead of ref?

Those are two independent schemas intended to describe the same notice type:
* The "Kafka" one: [fermi/gbm/FermiGBMAlert.schema.json](https://github.com/shb46/gcn-schema/blob/fermi/gcn/notices/fermi/gbm/FermiGBMAlert.schema.json)
   + Makes `$ref`-s to the core schema.
   + Doesn't `$ref` the strict schema. 
   + Validates new notices received by the Kafka server.
   + More permissive compared to the strict schema.


* The "strict" schema: [fermi/gbm/strict/FermiGBMAlert.schema.json](https://github.com/shb46/gcn-schema/blob/fermi/gcn/notices/fermi/gbm/FermiGBMAlert.schema.json):
   + Doesn't `$ref` the core schema, nor the "Kafka" schema. 
   + Is less permissive than the Kafka schema.
   + Clarifies what to expect in a notice, incl. which core fields are not used.
   + Used by the producer, in addition to the Kafka schema.

gcn/notices/fermi/gbm/FermiGBMAlert.schema.json
shb46: @Vidushi-GitHub: Please review the updated "description".

Vidushi-GitHub:
Hi Boyan, 
Could you please change the filename?
fermi/gcn/notices/fermi/gbm/FermiGBMAlert.schema.json to fermi/gcn/notices/fermi/gbm/alert.schema.json?

As it's already in fermi/gbm folder, again fermiGBM is redundant.

Vidushi-GitHub:
> > We have planned a mission page, where you can keep the content present at readme file.
> 
> * Can we have an instrument page as well?  the lat and the GBM instruments are run by separate teams.
> * Can notice type have its own page?

For Fermi mission page, we have information for GCN Classic Notices at https://gcn.nasa.gov/missions/fermi, and would move forward with updating it.
@jracusin shall we separate out GBM, LAT content into two mission pages as available at https://gcn.nasa.gov/missions/fermi Or, just update the existing one as single page?

Vidushi-GitHub:
> Those are two independent schemas intended to describe the same notice type:
> 
> * The "Kafka" one: [fermi/gbm/FermiGBMAlert.schema.json](https://github.com/shb46/gcn-schema/blob/fermi/gcn/notices/fermi/gbm/FermiGBMAlert.schema.json)
>   
>   * Makes `$ref`-s to the core schema.
>   * Doesn't `$ref` the strict schema.
>   * Validates new notices received by the Kafka server.
>   * More permissive compared to the strict schema.
> * The "strict" schema: [fermi/gbm/strict/FermiGBMAlert.schema.json](https://github.com/shb46/gcn-schema/blob/fermi/gcn/notices/fermi/gbm/FermiGBMAlert.schema.json):
>   
>   * Doesn't `$ref` the core schema, nor the "Kafka" schema.
>   * Is less permissive than the Kafka schema.
>   * Clarifies what to expect in a notice, incl. which core fields are not used.
>   * Used by the producer, in addition to the Kafka schema.
> 
> [Edit:] I should rename the strict schema file so it has a different name.

I am trying to understand why do you wanna create two schemas for same notice types? What is advantage of not referring $ref or kafka schema? Would you produce same notices from two schemas?
Definitions/clarifications shouldn't be problem, that can be added in alert schema as well. 

@lpsinger, @jracusin do you have comments on the same notices from different schemas, alert ($ref to core) and strict (own definitions)?

shb46:
@jracusin: 
Re. separate GBM/LAT pages.
Please let me know if it is possible.  If yes, we should get the consensus of the GBM and LAT teams before we make a change.  I think it will work with or without a "mission" page, i.e. either one of:
> - Missions
>>   - Fermi
>>>   - GBM
>>>   - LAT

or

> - Missions
>>   - FermiGBM
>>   - FermiLAT

shb46:
@Vidushi-GitHub: File renamed. The JPG and MD files removed.

shb46:
@jracusin: P.S.: I'm not planning to change the mission page at this time.

dakota002:
Hi @shb46 , I have a couple additional requests for this update:

Can you rename the strict schema and examples as well:
- `gcn/notices/fermi/gbm/strict/FermiGBMStrictAlert.schema.json` -> `gcn/notices/fermi/gbm/strict/alert.schema.json`
- `gcn/notices/fermi/gbm/strict/FermiGBMAlert1.example.json` -> `gcn/notices/fermi/gbm/strict/alert.FermiGBMAlert1.example.json`
- `gcn/notices/fermi/gbm/strict/FermiGBMAlert2.example.json` -> `gcn/notices/fermi/gbm/strict/alert.FermiGBMAlert2.example.json`
- `gcn/notices/fermi/gbm/strict/FermiGBMStrictAlert1.example.json` -> `gcn/notices/fermi/gbm/strict/alert.FermiGBMStrictAlert1.example.json`
- `gcn/notices/fermi/gbm/strict/FermiGBMStrictAlert2.example.json` -> `gcn/notices/fermi/gbm/strict/alert.FermiGBMStrictAlert2.example.json`

The schema name change reason is the same as the other schema as @Vidushi-GitHub mentioned, as the path already gives the name. As for the examples, the pattern of: `alert.[any specific name for the example].example.json` is recognized in our schema browser and parsers, and will help automate our docs. 

Once this is all set, could you then squash your commits down to 1 single commit? I can assist in this if you have questions before doing so.

shb46:
Hi, @dakota002,

Please check the new file names before squashing.

dakota002:
> Hi, @dakota002,
> 
> Please check the new file names before squashing.

The names look good to me!

shb46:
@dakota002: Did I squash it enough?

dakota002:
> @dakota002: Did I squash it enough?

I don't believe so, I see 52 commits now. You may have pulled the history back into your branch before pushing it out. You should just run the command `git push -f` to assert that the new history is correct

Vidushi-GitHub (changes requested):
It looks good to me. 

The only issue we need to address is the strict schema being used.
Would you like to create it for validation purposes only, or do you plan to have an additional topic for strict schema for the team's use?

shb46:
Hi, @Vidushi-GitHub: 

Regarding your latest comment.

> It looks good to me.

Let's not forget the "hitl" field before approving, in case we decide to drop it.

> The only issue we need to address is the strict schema being used. Would you like to create it for validation purposes only, or do you plan to have an additional topic for strict schema for the team's use?

The producer will validate against both the non-strict and the strict schemas at its end.  Kafka can use only the non-strict.  It also serves as a source-as-documentation.

jracusin:
> > > We have planned a mission page, where you can keep the content present at readme file.
> > 
> > 
> > 
> > * Can we have an instrument page as well?  the lat and the GBM instruments are run by separate teams.
> > * Can notice type have its own page?
> 
> For Fermi mission page, we have information for GCN Classic Notices at https://gcn.nasa.gov/missions/fermi, and would move forward with updating it. @jracusin shall we separate out GBM, LAT content into two mission pages as available at https://gcn.nasa.gov/missions/fermi Or, just update the existing one as single page?

No, we should modify the existing Fermi mission page, and can have sections for GBM & LAT.

jracusin:
> Hi, @Vidushi-GitHub:
> 
> Regarding your latest comment.
> 
> > It looks good to me.
> 
> Let's not forget the "hitl" field before approving, in case we decide to drop it.
> 
> > The only issue we need to address is the strict schema being used. Would you like to create it for validation purposes only, or do you plan to have an additional topic for strict schema for the team's use?
> 
> The producer will validate against both the non-strict and the strict schemas at its end. Kafka can use only the non-strict. It also serves as a source-as-documentation.

Each schema is for a topic, so you cannot produce multiple schema to the same topic.

gcn/notices/fermi/gbm/alert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+      "type": "string",
+      "description": "Notice title"
+    },
+    "hitl": {
+      "type": "boolean",
+      "description": "\"Human in the loop\".  When false, the notice was generated entirely automatically.  A value of true means that a human had an input, while the notice was being prepared.  The latter includes notices, which weren't altered by the human input, i.e. simply confirmed or approved by a human for further processing."
+    },
+    "trigger_algorithm_number": {
+      "type": "number",
+      "description": "The criterion number that caused the flight software to trigger. Not to be confused with flight software or any other version number."
```
jracusin, line 44: "Not to be confused with flight software or any other version number." - This doesn't really make sense.  Can you either list the corresponding energy range & timescale for that trigger_algorithm_number or refer to a table on the mission page that describes them?

gcn/notices/fermi/gbm/alert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    {
+      "$ref": "detectors.schema.json"
+    }
+  ],
+  "properties": {
+    "notice_type": {
+      "type": "string",
+      "description": "Notice title"
+    },
+    "hitl": {
```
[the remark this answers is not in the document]
jracusin: Could this instead be an enum with different types of localizations rather than a confusing boolean?  e.g. flight, ground, HITL?

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,174 @@
...
+  "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/strict/alert.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "GCN Schema for Fermi GBM Alert Notices",
+  "description": "This schema is a stricter version of the more permissive main schema, https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/alert.schema.json.  The Kafka server uses the main schema for new notices.  Each Alert notice (classic numeric type 110) must validate against the main schema and should also validate against this one.  Examples in this directory come in two versions referencing the main and the strict schemas respectively.  The intent of the strict schema is to help understanding the Alert notice, since it better describes what content to be expected.",
+  "type": "object",
+  "properties": {
+    "$schema": {
+      "enum": [
+        "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/alert.schema.json",
+        "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/strict/alert.schema.json"
```
jracusin, line 11: This shouldn't reference a particular version of the schema, but should be a relative reference.

gcn/notices/fermi/gbm/alert.schema.json
shb46: The trigger energy range and timescale are already included in the Alert notice.

gcn/notices/fermi/gbm/alert.schema.json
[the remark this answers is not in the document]
shb46: @jracusin: Then we would be more stright forward to name it "human localization". HITL Is more general. I don't mind dropping it from the Alert. It should be in the Core anyway.

gcn/notices/fermi/gbm/strict/alert.schema.json
dakota002: The exact value here should actually be:

```json
"$schema": { "type": "string", "format": "uri" }
```
The only purpose for defining the `$schema` attribute in the properties section is so that, when used with `'unevaluatedProperties': false`, the validator knows that there should be a `$schema` in a notice

gcn/notices/fermi/gbm/alert.schema.json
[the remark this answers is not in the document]
Vidushi-GitHub: After discussions "HILT" specific to GBM is fine. 
Style looks inconsistent to other schema fields, if possible make the property name explicit like, human_in_loop.

@shb46 are you considering to put it as enum?

Let's move this forward to some conclusion.

gcn/notices/fermi/gbm/detectors.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "type": "object",
+  "properties": {
+    "detectors_status": {
+      "type": "array",
+      "items": {
+        "$ref": "../../core/DetectorStatus.schema.json"
+      },
+      "minItems": 14,
+      "maxItems": 14,
+      "description": "Triggered detectors"
```
Vidushi-GitHub, line 15: Please expand it

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,168 @@
...
+      "default": "Fermi",
+      "description": "Name of Mission or Telescope reporting the event"
+    },
+    "notice_type": {
+      "const": "FERMI_GBM_ALERT",
+      "description": "Notice title"
+    },
+    "rate_duration": {
+      "type": "number",
+      "description": "Duration in seconds of the ime interval used in the signal to noise ratio calculation"
```
Vidushi-GitHub, line 107:
```suggestion
      "description": "Duration in seconds of the time interval used in the signal to noise ratio calculation"
```

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,168 @@
...
+      "description": "Incremental number for notices related to the same trigger (ex: 1, 2, 3)"
+    },
+    "spectrum": {
+      "const": "energy",
+      "default": "energy",
+      "description": "high-energy or optical or radio observations, if not parsed, then default band is energy"
+    },
+    "trigger_algorithm_number": {
+      "type": "number",
+      "description": "The criterion number that caused the flight software to trigger. Not to be confused with flight software or any other version number."
```
Vidushi-GitHub, line 133: make it consistent with above definition.

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,168 @@
...
+    },
+    "trigger_type": {
+      "const": "rate",
+      "default": "rate",
+      "description": "Type of trigger algorithm used to identify the transient event."
+    },
+    "units": {
+      "const": "keV",
+      "default": "keV",
+      "description": "Units of band range, if not parsed, then default energy is keV"
```
Vidushi-GitHub, line 147:
```suggestion
      "description": "Units of energy band range, if not parsed, then default energy is keV"
```

gcn/notices/fermi/gbm/strict/alert.2strict.example.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/strict/alert.schema.json",
+  "additional_info": "*** ABOUT THIS EXAMPLE: This is an example of a Fermi GBM Alert GCN notice. Only stricly necessary fields are present. This 'about' note is not going to be part of a real-life notice. This field may be missing or empty in case of no comments. An example of a typical notice comment follows. ***\n\nLightcurve and other quicklook files will not be created until about 15 minutes after the trigger.",
```
Vidushi-GitHub, line 3:
```suggestion
  "additional_info": "*** ABOUT THIS EXAMPLE: This is an example of a Fermi GBM Alert GCN notice. Only stricly necessary fields are present. This 'about' note is not going to be part of streamed/real notices. This field may be missing or empty in case of no comments. An example of a typical notice comment follows. ***\n\nLightcurve and other quicklook files will not be created until about 15 minutes after the trigger.",
```

gcn/notices/fermi/gbm/strict/alert.2.example.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/alert.schema.json",
+  "additional_info": "*** ABOUT THIS EXAMPLE: This is an example of a Fermi GBM Alert GCN notice. Only stricly necessary fields are present. This 'about' note is not going to be part of a real-life notice. This field may be missing or empty in case of no comments. An example of a typical notice comment follows. ***\n\nLightcurve and other quicklook files will not be created until about 15 minutes after the trigger.",
```
Vidushi-GitHub, line 3:
```suggestion
  "additional_info": "*** ABOUT THIS EXAMPLE: This is an example of a Fermi GBM Alert GCN notice. Only stricly necessary fields are present. This 'about' note is not going to be part of streamed notices. This field may be missing or empty in case of no comments. An example of a typical notice comment follows. ***\n\nLightcurve and other quicklook files will not be created until about 15 minutes after the trigger.",
```

gcn/notices/fermi/gbm/alert.schema.json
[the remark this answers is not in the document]
shb46: @Vidushi-GitHub: I renamed the field and adjusted the description accordingly. I'm keeping it as a boolean.

gcn/notices/fermi/gbm/detectors.schema.json
shb46: @Vidushi-GitHub: Description improved.

shb46: @Vidushi-GitHub: Typo fixed. Thanks for the catch!

gcn/notices/fermi/gbm/strict/alert.schema.json
shb46: @Vidushi-GitHub: Typo fixed. Thanks for the catch!

gcn/notices/fermi/gbm/strict/alert.schema.json
shb46: @Vidushi-GitHub: Replaced "band range" with "rate_energy_range", the explicit field name. It wasn't clear. Thanks for pointing this out.

gcn/notices/fermi/gbm/strict/alert.2strict.example.json
shb46: @Vidushi-GitHub: Done.

gcn/notices/fermi/gbm/strict/alert.2.example.json
shb46: @Vidushi-GitHub: Done. Thanks for all the great suggestions!

gcn/notices/fermi/gbm/strict/alert.1.example.json
```diff
@@ -0,0 +1,43 @@
...
+    "on",
+    "on",
+    "on",
+    "on",
+    "on",
+    "on",
+    "on"
+  ],
+  "event_name": ["GRB240824A"],
+  "hitl": false,
```
Vidushi-GitHub, line 25: Please make the example field name consistent with schema.
Additionally, validate the schema against examples, in case any change is left out.

shb46: @Vidushi-GitHub: Changed. Thank you for the reminder!

Vidushi-GitHub (approved):
@jracusin please have a look, and provide your feedback/approval.

gcn/notices/fermi/gbm/strict/alert.schema.json
Vidushi-GitHub: @shb46 reminder.

shb46: @Vidushi-GitHub: Done. Thanks for the reminder.

gcn/notices/fermi/gbm/alert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+      "$ref": "../../core/Statistics.schema.json"
+    },
+    {
+      "$ref": "detectors.schema.json"
+    }
+  ],
+  "properties": {
+    "notice_type": {
+      "type": "string",
+      "description": "Notice title"
```
jracusin, line 36: Can this be an enum?  How many options are there for notice title?  Are you always going to refer back to GCN Classic notice types?  You don't need to be tied to those.

gcn/notices/fermi/gbm/alert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    "human_in_the_loop": {
+      "type": "boolean",
+      "description": "Someimes abbreviated as \"HITL\".  When false, the notice was generated entirely automatically.  A value of true means that a human had an input, while the notice was being prepared.  The latter includes notices, which weren't altered by the human input, but a human was given the option to make an input, or asked to confirm, or approve before further processing."
```
jracusin, line 40: Is HITL also a notice type?  This description could also be much more concise. For example:
"True when a human had input or confirmed before further processing; False when the process was fully automated."

gcn/notices/fermi/gbm/alert.schema.json
```diff
@@ -0,0 +1,51 @@
...
+    },
+    "trigger_algorithm_number": {
+      "type": "number",
+      "description": "The number identifying the algorithm that detected the gamma-ray burst and caused the trigger.  \"Trigger algorithm\" here is a moniker for a method or a criterion that watches for significant rate increase in the observed rate.  Multiple \"trigger algorithms\" are active simultaneously and any one can trigger independently.  The trigger_algorithm_number, tells which one did."
```
jracusin, line 44: Are these trigger algorithms described somewhere in a document or paper. Can you just reference that document with something like the following.  There's too much extra information in this description.
"The instrument flight software algorithm that triggered on this event, dependent on timescale, energy, and significance as described in ...".

gcn/notices/fermi/gbm/alert.example.json
```diff
@@ -0,0 +1,43 @@
...
+    "on",
+    "on",
+    "on"
+  ],
+  "event_name": ["GRB240824A"],
+  "human_in_the_loop": false,
+  "id": ["745121685", "bn240812094"],
+  "instrument": "GBM",
+  "latitude": 23.83,
+  "light_curve_url": "http://heasarc.gsfc.nasa.gov/FTP/fermi/data/gbm/triggers/2024/bn240812094/quicklook/glg_lc_medres34_bn240812094.gif",
```
jracusin, line 29: These don't need to be alphabetical, and it seems strange to separate latitude and longitude.  In general, these examples should be in some logical order.

gcn/notices/fermi/gbm/detectors.schema.json
```diff
@@ -0,0 +1,18 @@
...
+  "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/detectors.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "",
+  "description": "",
+  "type": "object",
+  "properties": {
+    "detectors_status": {
+      "type": "array",
+      "items": {
+        "$ref": "../../core/DetectorStatus.schema.json"
```
jracusin, line 11: These can be referenced by name rather than indexed, which would then not require this detectors schema at all.  See https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/burstcube/alert.example.json or https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/glowbug/Alert.schema.json, which you previously came up with.

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,168 @@
...
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "Name of the event, for example GRB170817A"
+    },
+    "human_in_the_loop": {
+      "type": "boolean",
+      "default": false,
+      "description": "Someimes abbreviated as \"HITL\".  When false, the notice was generated entirely automatically.  A value of true means that a human had an input, while the notice was being prepared.  The latter includes notices, which weren't altered by the human input, but a human was given the option to make an input, or asked to confirm, or approve before further processing."
```
jracusin, line 58:
```suggestion
      "description": "Sometimes abbreviated as \"HITL\".  When false, the notice was generated entirely automatically.  A value of true means that a human had an input, while the notice was being prepared.  The latter includes notices, which weren't altered by the human input, but a human was given the option to make an input, or asked to confirm, or approve before further processing."
```

gcn/notices/fermi/gbm/strict/alert.schema.json
```diff
@@ -0,0 +1,168 @@
+{
+  "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/strict/alert.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "GCN Schema for Fermi GBM Alert Notices",
+  "description": "This schema is a stricter version of the more permissive main schema, https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/alert.schema.json.  The Kafka server uses the main schema for new notices.  Each Alert notice (classic numeric type 110) must validate against the main schema and should also validate against this one.  Examples in this directory come in two versions referencing the main and the strict schemas respectively.  The intent of the strict schema is to help understanding the Alert notice, since it better describes what content to be expected.",
```
jracusin, line 5: I still don't understand the purpose or use of this strict schema.  Each schema is a Kafka topic, though you are welcome to multiple examples.  The GBM topic will be gcn.notices.fermi.gbm.alert, so what is the use case for gcn.notices.fermi.gbm.strict.alert?

gcn/notices/fermi/gbm/strict/alert.2.example.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/fermi/gbm/alert.schema.json",
+  "additional_info": "*** ABOUT THIS EXAMPLE: This is an example of a Fermi GBM Alert GCN notice. Only stricly necessary fields are present. This 'about' note is not going to be part of streamed/real notices. This field may be missing or empty in case of no comments. An example of a typical notice comment follows. ***\n\nLightcurve and other quicklook files will not be created until about 15 minutes after the trigger.",
```
jracusin, line 3: What is different between these examples, please add some info about their difference rather than number them in their name.

gcn/notices/fermi/gbm/alert.schema.json
shb46: Added `"const": "FERMI_GBM_ALERT"`.  This is redundant with `$schema` and the topic.  Thanks for pointing out.

gcn/notices/fermi/gbm/alert.schema.json
shb46: Shorter and better.  Thanks for the suggestion.

gcn/notices/fermi/gbm/alert.example.json
shb46: Alphabetical order helps bookkeeping. I'd like to keep it this way.

gcn/notices/fermi/gbm/detectors.schema.json
shb46: I think we are discussing two design points here:

1. Array vs. names;
2. To `$ref` or not to `$ref` the `core/DetectorStatus`.

Thanks for suggesting an alternative to the array. The two choices were discussed at our pipeline meeting (1 vote for names) and the IOTL meeting. The 14-array won. Consider that it is a bit mask in the data products, in which the bits are indexed 0-13.  The same is in the classic notes, incl. the text ones. Also the old and the new GBM tools use straight arrays. In retrospect, I should had used an array for Glowbug as well.

On point (2), if you would like to remove the `$ref`, I'll be happy to replace it with the `enum`. just let me know.

gcn/notices/fermi/gbm/strict/alert.schema.json
shb46: Fixed. Thanks for catching it.

gcn/notices/fermi/gbm/strict/alert.schema.json
shb46: I wasn't planning on having a `gcn.notices.fermi.gbm.strict.alert` topic, only the `gcn.notices.fermi.gbm.alert`. The same way the server functionality doesn't depend on the examples, it should ignore the strict schemas. If the server does notice validation, it should use only the `.../gbm/alert.schema.json`. All I want the strict schema to do, is will just be there in addition to the examples and will serve a similar illustrative/documentation purpose.

The JSON objects described by the strict schema are a subset of the objects described by the "non-sctrict" one. If, by looking at the examples, one has doubts about what to expect, then the strict schema will help clear it up. This will become more so with the other notice types.

gcn/notices/fermi/gbm/strict/alert.2.example.json
shb46: Agree and done. Thanks!

shb46:
@Vidushi-GitHub: Are we ready to close this PR? Who is going to close it?

Vidushi-GitHub:
> @Vidushi-GitHub: Are we ready to close this PR? Who is going to close it?

No, we need approvals for latest feedbacks. I will remind Judy.

gcn/notices/fermi/gbm/strict/alert.schema.json
lpsinger: @shb46, I looked at a diff between gcn/notices/fermi/gbm/alert.schema.json and gcn/notices/fermi/gbm/strict/alert.schema.json. It looks like the only difference is that in the former, you include types from core schema, whereas in the latter, you hardcode the definitions of the fields from the core schema that you are actually using. Please don't do that! Please remove the "strict" schema directory from your PR.

I think that this exposes an interesting issue, that when you include a core schema, it is not clear to users which of its many optional fields you actually intend to populate. That affects all GCN notice types, not just Fermi. @jracusin, let's open another issue to study that problem and come up with options.

shb46: Done.

gcn/notices/fermi/gbm/detectors.schema.json
Vidushi-GitHub: Hi @shb46, Could you please implement detectors schema within `alert.schema.json`?
I understand the logic of making it as a library and calling it. But it's not a few lines field and would make `alert.schema.json` look a self consistent schema.

Are you planning to add more topics in future and wanna call `detectors.schema.json`?

shb46: @Vidushi-GitHub: The detectors schema, if we keep it, will be shared with another topic. Do you still want me to make it inline?

Vidushi-GitHub: @shb46 Ok, thanks. If you plan to have more topics related to Fermi/GBM, then that's fine.
Please remember to limit topics per mission to simplify user subscriptions to new Notices. The new Schema allows multiple Notices using a single Schema since fields are not mandatory.

shb46: @Vidushi-GitHub: Our intent was for GBM to have 5 topics.  Hope that is not too much.  I suspect we might need a 6th one.  Separately, there might be topics for internal distribution only but this hasn't been discussed yet.

My plan is to have a 1-to-1-to-1 correspondence of classic notice types, schemas and topics. I think it is a little too late to steer towards making a single schema. We are ahead with approving the ALERT schema and it will set us back. My deadline is Dec 12.

jracusin: @shb46 The design of the unified schema has always been to incorporate many of the classic notice types into the same topic so users can only follow 1 or 2 topics per instrument.  You can send updates as more information to fill out as the schema becomes available, but rebuilding GCN Classic in the JSON-format was never the intention. We should discuss the deadline offline, as I assume that's tied to senior review.

shb46: @jracusin: No problem, I can use topic `gcn.notices.fermi.gbm` for everything.

PR #240, opened 2025-02-05
Core Schema: Notice_type

Vidushi-GitHub:
# Description

Single schema is requested to handle multiple notice types from the mission teams. 

While the notice types are distinct in the example files, in the streaming process, all notices are generated from the same pipeline.
This property is useful to distinct the types of notices. 

Also, 'notice_type' is consistent with GCN Classic Notices.

### DIFF
--- a/gcn/notices/core/Reporter.schema.json
+++ b/gcn/notices/core/Reporter.schema.json
@@ -13,6 +13,10 @@
       "type": "string",
       "description": "Name of the Instrument reporting the event"
     },
+    "notice_type": {
+      "type": "string",
+      "description": "Defines the category or type of notice and serves as its title. It is recommended when a single schema is used to produce multiple notice types."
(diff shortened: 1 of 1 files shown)


jracusin (approved):

jracusin:
We decided to instead keep this as mission specific so that they can use custom enums, so it will not be part of the core schema.  We will also add documentation to the unified schema page with this guidance.  Issue #2878

Vidushi-GitHub:
> We decided to instead keep this as mission specific so that they can use custom enums, so it will not be part of the core schema. We will also add documentation to the unified schema page with this guidance. Issue #2878

We decided "alert_type" but it has degeneracy with alert schema: https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/core/Alert.schema.json

PR #236, opened 2025-01-22
Common neutrino telescope alert schema

blaufuss:
# Description
This PR adds a common GCN alert schema for alerts from neutrino telescopes.  

It also reimplements the IceCube Gold/Bronze track alerts using this common schema, with example.

KM3Net will also implement their alerts using this common base schema.  

We are happy to move the location of the shared base schema as determined by GCN dev team.

PR #243, opened 2025-02-12
Add templates folder and sample schema

Vidushi-GitHub:
# Description
Template folder for hosting common GCN alert schema, 
which can be utilized by similar type of observatories example neutrino observations. 

Also, contains a sample schema and example for what schema should contain.

# Related Issue(s)
PR #236

### DIFF
--- a/gcn/notices/templates/sample.example.json
+++ b/gcn/notices/templates/sample.example.json
@@ -1,0 +1,8 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/templates/sample.schema.json",
+  "alert_datetime": "2023-09-28T01:40:30Z",
+  "ra": 197.44871198,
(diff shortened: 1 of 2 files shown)


Vidushi-GitHub:
> I don't get it. Why do we need a templates directory, when we have so many examples of schema already?

Neutrino community, right now IceCube and KM3Net (refer #236  and KM3Net email), has decided to adopt a generic unified schema template for neutrino observations. They would like to keep it as a reference in the gcn-schema directory.

Additionally, I plan to provide a minimal schema example (will update for: alert_type, id, trigger_time, ra, dec, etc) for upcoming missions. This was a feedback during the AstroCOLIBIRI meeting.

Vidushi-GitHub:
As discussed, not required, will add a minimal schema in the document.

PR #261, opened 2025-04-11
Update ID for single string and array

Vidushi-GitHub:
# Description
Making ID consistent with single number and array both

### DIFF
--- a/gcn/notices/core/Event.schema.json
+++ b/gcn/notices/core/Event.schema.json
@@ -11,9 +11,17 @@
       "description": "Name of the event (ex: GRB 170817A, GW170817, AT2017gfo, SSS 17a)"
     },
     "id": {
-      "type": "array",
-      "items": { "oneOf": [{ "type": "string" }, { "type": "number" }] },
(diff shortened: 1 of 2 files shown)


gcn/notices/core/Event.schema.json
```diff
@@ -11,9 +11,17 @@
       "description": "Name of the event (ex: GRB 170817A, GW170817, AT2017gfo, SSS 17a)"
     },
     "id": {
-      "type": "array",
-      "items": { "oneOf": [{ "type": "string" }, { "type": "number" }] },
-      "description": "Instrument-specific trigger ID (ex: bn230313485 (Fermi), 1159327 (Swift)) or alternate ID"
+      "oneOf": [
+        {
+          "type": "string",
```
jracusin, line 16: This should be a string or number like the array to be consistent.

Vidushi-GitHub: Done

jracusin: Sorry, we decided to go with just strings for simplicity for future databases.

PR #217, opened 2024-11-05
FRB Schema for the CHIME Telescope

tabbott36:
# Description
Added a schema for fast radio bursts (FRBs). I have also included an example json for FRB alerts from the Canadian Hydrogen Intensity Mapping Experiment (CHIME) for each alert type (detection, subsequent, retraction, and update). 

# Questions for GCN Team
- Can I use a Transient Name Server name for the "id" parameter in the Event schema?

# Testing
- I modelled the schema after other observatories' schemas but am open to suggestions regarding testing.

PR #295, opened 2025-08-15
DSA110: SNR and Unevaluated Property

Vidushi-GitHub:
# Description
Please keep the dsa110 schema consistent with CHIME definitions.

I understand that you kept the SNR property earlier, 
but we suggested `rate_snr`, now found the reason why it should be `snr` instead of `rate_snr`.
Please check: https://github.com/nasa-gcn/gcn-schema/pull/217#discussion_r1951284867

Additionally, some properties are not defined in the schema but are present in the examples
Thus, I have added "unevaluatedProperties": false so that all undefined fields are captured.
Please define those properties in your schema.


caseyjlaw (commented):
Why was the `rate_duration` field removed from the example? The duration (length of signal in time) is a standard measurement that consumers will be interested in.

Vidushi-GitHub:
> Why was the `rate_duration` field removed from the example? The duration (length of signal in time) is a standard measurement that consumers will be interested in.

I understand it should be included in FRB Notices. As, `rate_snr` isn’t there; it’s not an appropriate property. 
We need to come up with a common definition for both CHIME and DSA110 notices, either `sampling_time` or `duration`.
Thanks to your PR, we found a missing definition `sampling_time` in CHIME @tabbott36, and will start a PR for the same definition.

caseyjlaw:
Ok, thanks for explaining. I'll wait for the feedback from CHIME. The `sampling_time` would be an instrumental parameter, fwiw.
I am comfortable using `rate_snr` and `rate_duration`, but still find the `rate_` part a bit hard to parse. Assuming `snr` and `duration` are not sufficiently precise, I'd lean toward a prefix or suffix of `event_` or `time_`.

Vidushi-GitHub:
> Ok, thanks for explaining. I'll wait for the feedback from CHIME. The `sampling_time` would be an instrumental parameter, fwiw. I am comfortable using `rate_snr` and `rate_duration`, but still find the `rate_` part a bit hard to parse. Assuming `snr` and `duration` are not sufficiently precise, I'd lean toward a prefix or suffix of `event_` or `time_`.

@caseyjlaw as duration would be first definition, please start modifying your schema. For now, may be add `event_duration`. 
Thanks for `sampling_time` information.

Vidushi-GitHub:
A new PR is created by the team: https://github.com/nasa-gcn/gcn-schema/pull/297
Thus closing it.

PR #307, opened 2025-09-18
Update units name

### LINKED ISSUE #305: CHIME/FRB: Update spectral band units
# Description

We realized a bug in property naming of `units` for the `spectral_band`.
It's is very generic name, should be self-explanatory.
Thus, we have tagged the core schema version to 5.0.0, which will break the CHIME/FRB schema.

Please update the schema property `units` to `spectral_band_units`.

Vidushi-GitHub:
# Description
Recently we have revised the name of core schema property.

`units` to `spectral_band_units`

Resolves #305


tabbott36 (approved):
Looks good. Thanks, Vidushi!

PR #324, opened 2025-12-05
Boom notice schema

antoine-le-calloch:
# Description
This PR adds the JSON schema and example for optical alerts with crossmatched GCN events coming from Boom.

# Related Issue(s)
This PR addresses part of the following issue: https://github.com/nasa-gcn/gcn.nasa.gov/issues/3328

# Testing
The files were successfully validated using the project's configured npm hooks, including Prettier formatting, ESLint checks, and JSON Schema validation

PR #343, opened 2026-03-03
Add Magnitude-based Schema

Vidushi-GitHub:
# Description

For optical surveys, we are adding magnitude-based schema.

It includes:

- Filter
- Magnitude
- Magnitude Error
- Magnitude System
- Limiting Magnitude
- Limiting Magnitude Sigma Value


tylerbarna (commented):
do we want to specify the allowed magnitude systems? I believe there was a list of allowed systems in the proposed schema

tylerbarna (approved):

Vidushi-GitHub:
> > > Do we need to specific the time at which the object was at this magnitude? Or would that be in observation_start, observation_stop?
> > 
> > 
> > Yes, it should be observation start, stop. However, I noticed BOOM is using only observation_start. (https://github.com/nasa-gcn/gcn-schema/pull/324/changes)
> > If we still wish to add, we can add mid time?
> 
> We need start/stop or or mid with error bars, or start + exposure.

Looking into ztf-style schema, they only use start of exposure (https://github.com/ZwickyTransientFacility/ztf-avro-alert/blob/master/schema/prv_candidate.avsc) 
BOOM also uses observation_start

DateTime.schema.json as one of the blocks seems fine, rather than defining the same fields multiple times. 
It includes observation_livetime; for completeness, we could add exposure_time there. What do you think?

jracusin:
> > > > Do we need to specific the time at which the object was at this magnitude? Or would that be in observation_start, observation_stop?
> > > 
> > > 
> > > Yes, it should be observation start, stop. However, I noticed BOOM is using only observation_start. (https://github.com/nasa-gcn/gcn-schema/pull/324/changes)
> > > If we still wish to add, we can add mid time?
> > 
> > 
> > We need start/stop or or mid with error bars, or start + exposure.
> 
> Looking into ztf-style schema, they only use start of exposure (https://github.com/ZwickyTransientFacility/ztf-avro-alert/blob/master/schema/prv_candidate.avsc) BOOM also uses observation_start
> 
> DateTime.schema.json as one of the blocks seems fine, rather than defining the same fields multiple times. It includes observation_livetime; for completeness, we could add exposure_time there. What do you think?

Okay, we shouldn't add new requirements, so we can just encourage use of photometry with observation_start or date time so there's a time associated.

jracusin (approved):

tylerbarna (approved):

tylerbarna (changes requested):

tylerbarna (approved):

PR #344, opened 2026-03-05
Fix photometry schema filename to Uppercase

Vidushi-GitHub:
# Rename photometry.schema.json to Photometry.schema.json

Core schema follows uppercase naming conventions according to SnakeCase scheme.
By mistake I added photometry.schema.json instead of Photometry.schema.json.


tylerbarna (changes requested):
I believe photometry.schema.json needs to be deleted, no?

tylerbarna (approved):

PR #352, opened 2026-04-08
feat: add SNEWS2 coincidence alert schema

medhansh29:
## Summary
Adds the initial JSON schema for Supernova Early Warning System 2.0 (SNEWS2) coincidence tier alerts to the GCN schema repository.

## Files Added
- `gcn/notices/snews2/coincidence.schema.json` — JSON schema defining the SNEWS2 CoincidenceTier alert format
- `gcn/notices/snews2/coincidence.test_alert.example.json` — Example payload conforming to the schema

## Schema Design
The schema inherits from 5 GCN Core schemas using `allOf`:
- `core/Alert.schema.json`
- `core/Event.schema.json`
- `core/Reporter.schema.json`
- `core/DateTime.schema.json`
- `core/Statistics.schema.json`

SNEWS2-specific fields include:
- `snews2_tier` — The SNEWS tier (CoincidenceTier only at this stage)
- `detector_names` — Array of contributing detector names
- `event_times_utc` — Array of neutrino arrival times per detector
- `tier_data` — Tier-specific payload (p_values, false_alarm_prob, etc.)

## Notes
- Only the `CoincidenceTier` alert type is published to GCN at this time
- All tests pass with `npm run test`

### DIFF
--- a/gcn/notices/snews2/coincidence.real_alert_no_triangulation.example.json
+++ b/gcn/notices/snews2/coincidence.real_alert_no_triangulation.example.json
@@ -1,0 +1,24 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/snews2/coincidence.schema.json",
+  "alert_datetime": "2026-06-18T16:23:21.937Z",
+  "alert_tense": "current",
+  "alert_type": "initial",
+  "id": "SNEWS_Coincidence_ALERT 2026-06-18T16:23:21.937Z",
+  "mission": "SNEWS",
+  "messenger": "Neutrino",
+  "record_number": 1,
+  "trigger_time": "2026-06-18T16:21:07.312Z",
+  "far": 0.0001,
+  "snews_schema_version": "1.0",
+  "snews2_tier": "CoincidenceTier",
+  "detector_status": {
+    "Super-K": "triggered",
+    "Borexino": "triggered"
+  },
+  "event_times_utc": ["2026-06-18T16:21:07.312Z", "2026-06-18T16:21:07.318Z"],
+  "tier_data": {
+    "neutrino_time_utc": "2026-06-18T16:21:07.312000Z",
+    "p_values": [0.01, 0.09],
+    "false_alarm_prob": 0.0001
+  }
+}
--- a/gcn/notices/snews2/coincidence.schema.json
+++ b/gcn/notices/snews2/coincidence.schema.json
@@ -1,0 +1,64 @@
+{
+  "$id": "https://gcn.nasa.gov/schema/main/gcn/notices/snews2/coincidence.schema.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "title": "SNEWS2 Coincidence Alert",
+  "description": "Supernova Early Warning System (SNEWS 2.0) coincidence network capabilities.",
+  "allOf": [
+    { "$ref": "../core/Alert.schema.json" },
+    { "$ref": "../core/Event.schema.json" },
+    { "$ref": "../core/Reporter.schema.json" },
+    { "$ref": "../core/DateTime.schema.json" },
+    { "$ref": "../core/Statistics.schema.json" }
+  ],
+  "properties": {
+    "snews_schema_version": {
+      "type": "string",
+      "description": "SNEWS2 schema version."
+    },
+    "snews2_tier": {
+      "type": "string",
+      "description": "The categorization of the SNEWS 2.0 alert type (e.g., CoincidenceTier, SignificanceTier, TimingTier) indicating the nature of the supernova detection."
+    },
+    "detector_status": {
+      "type": "object",
+      "description": "Status of SNEWS2 detectors at the time of the alert.",
+      "additionalProperties": { "$ref": "../core/DetectorStatus.schema.json" }
+    },
+    "event_times_utc": {
+      "type": "array",
+      "items": { "$ref": "../core/DateTime.schema.json#/properties/trigger_time" },
+      "description": "Primary timestamp(s) for the event from the detectors."
+    },
+    "tier_data": {
+      "type": "object",
+      "description": "Tier-specific data payload containing p-values, detector statistics, etc.",
+      "properties": {
+        "neutrino_time_utc": {
+          "type": "string",
+          "description": "The aggregated neutrino arrival time across the coincidence network."
+        },
+        "p_values": {
+          "type": "array",
+          "items": {
+            "type": "number",
+            "minimum": 0
+          },
+          "description": "The p-values representing the significance of the signal from each respective detector."
+        },
+        "false_alarm_prob": {
+          "type": "number",
+          "description": "The calculated false alarm probability of the coincidence."
+        }
+      },
+      "additionalProperties": true
+    }
+  },
+  "required": [
+    "snews_schema_version",
+    "snews2_tier",
+    "detector_status",
+    "event_times_utc",
+    "tier_data"
+  ]
+}
--- a/gcn/notices/snews2/coincidence.schema.json
+++ b/gcn/notices/snews2/coincidence.schema.json
@@ -5,11 +5,24 @@
   "title": "SNEWS2 Coincidence Alert",
   "description": "Supernova Early Warning System (SNEWS 2.0) coincidence network capabilities.",
   "allOf": [
-    { "$ref": "../core/Alert.schema.json" },
-    { "$ref": "../core/Event.schema.json" },
-    { "$ref": "../core/Reporter.schema.json" },
-    { "$ref": "../core/DateTime.schema.json" },
-    { "$ref": "../core/Statistics.schema.json" }
+    {
+      "$ref": "../core/Alert.schema.json"
+    },
+    {
+      "$ref": "../core/Event.schema.json"
+    },
+    {
+      "$ref": "../core/Reporter.schema.json"
+    },
+    {
+      "$ref": "../core/DateTime.schema.json"
+    },
+    {
+      "$ref": "../core/Statistics.schema.json"
+    },
+    {
+      "$ref": "../core/Localization.schema.json"
+    }
   ],
   "properties": {
     "snews_schema_version": {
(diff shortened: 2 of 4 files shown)


gcn/notices/snews2/coincidence.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    { "$ref": "../core/Alert.schema.json" },
+    { "$ref": "../core/Event.schema.json" },
+    { "$ref": "../core/Reporter.schema.json" },
+    { "$ref": "../core/DateTime.schema.json" },
+    { "$ref": "../core/Statistics.schema.json" }
+  ],
+  "properties": {
+    "schema_version": {
+      "type": "string",
+      "description": "SNEWS2 schema version."
```
jracusin: Is this an internal SNEW2 tracking of your schema version or the tagged version of the GCN schema?  If the latter, then this gets specified in production with the schema keyword.

If it's an internal SNEWS2 schema version, then we should rename this to indicate that.

jracusin: For example:

   '$schema': 'https://gcn.nasa.gov/schema/vX.Y.Z/gcn/notices/mission/SchemaName.schema.json',

gcn/notices/snews2/coincidence.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    "snews2_tier": {
+      "type": "string",
+      "description": "The tier of the original SNEWS message."
```
jracusin: Can you be more descriptive about a tier? This could also be described in more detail on the mission page.

gcn/notices/snews2/coincidence.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    "detector_names": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "Originating detector(s) contributing to the coincidence."
+    },
+    "event_times_utc": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "Primary timestamp(s) for the event from the detectors."
```
jracusin, line 48: Is the event the first neutrino from each of the detectors?

jracusin: This could also be constructed as an array of datetime.trigger_time from the core schema

gcn/notices/snews2/coincidence.test_alert.example.json
```diff
@@ -0,0 +1,24 @@
...
+  "messenger": "Neutrino",
+  "trigger_time": "2026-04-03T16:21:07.312Z",
+  "far": 0.0001,
+  "schema_version": "1.0",
+  "snews2_tier": "CoincidenceTier",
+  "detector_names": ["Super-K", "IceCube", "KamLAND"],
+  "event_times_utc": [
+    "2026-04-03T16:21:07.312+00:00",
+    "2026-04-03T16:21:07.312+00:00",
+    "2026-04-03T16:21:07.312+00:00"
```
jracusin: What does the "+00:00" indicate? The format before this is the ISO1806 standard.

gcn/notices/snews2/coincidence.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    },
+    "snews2_tier": {
+      "type": "string",
+      "description": "The tier of the original SNEWS message."
+    },
+    "detector_names": {
+      "type": "array",
+      "items": { "type": "string" },
+      "description": "Originating detector(s) contributing to the coincidence."
+    },
```
jracusin: Have you considered using [DetectorStatus](https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/core/DetectorStatus.schema.json) from the core schema?  We could modify that to add another option than "triggered". This way you could list all the SNEWS2 detectors, which ones were observing at the time, and which ones detected the neutrinos?
A good example of this usage is [BurstCube](https://github.com/nasa-gcn/gcn-schema/blob/main/gcn/notices/burstcube/alert.example.json).

gcn/notices/snews2/coincidence.schema.json
medhansh29: This field tracks our internal SNEWS2 schema version. I'll make the change to use the correct way to name internal schema versions.

gcn/notices/snews2/coincidence.schema.json
medhansh29: In the SNEWS 2.0 network, a 'tier' categorizes the type of observation or alert generated by the detectors. The current tier we have opened up is the CoincidenceTier: A temporal coincidence of signals across multiple detectors.

medhansh29: I'll update the description in the schema too!

gcn/notices/snews2/coincidence.schema.json
medhansh29: Yes. Internally within SNEWS, event_times_utc (or neutrino_times_utc) corresponds to the time of the first detected neutrino from each respective detector in the coincidence.

Taken a look at the the datetime.trigger_time field in DateTime.schema.json and will implement it as $ref. Thank you!

gcn/notices/snews2/coincidence.test_alert.example.json
medhansh29: The +00:00 indicates the timezone offset, meaning the time is in UTC.

medhansh29:
Key Changes:

Schema Versioning: Renamed schema_version to snews_schema_version to completely eliminate any ambiguity with the standard JSON $schema keyword.

Detector Status Enrichment: Replaced the simple detector_names string array with a detector_status object pattern mapping to DetectorStatus.schema.json via additionalProperties. This matches patterns established by missions like BurstCube and allows tracking whether detectors were triggered, on, or off.

Core DateTime Referencing: Updated the event_times_utc array to explicitly use a JSON Pointer $ref to ../core/DateTime.schema.json#/properties/trigger_time. By doing this, we seamlessly inherit the core schema's documentation and standard ISO 8601 constraints without duplicating logic.

jracusin:
Is there a formal scheme for naming these SNEWS events with an `event_name`?

gcn/notices/snews2/coincidence.schema.json
```diff
@@ -0,0 +1,64 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "type": "object",
+  "title": "SNEWS2 Coincidence Alert",
+  "description": "Supernova Early Warning System (SNEWS 2.0) coincidence network capabilities.",
+  "allOf": [
+    { "$ref": "../core/Alert.schema.json" },
+    { "$ref": "../core/Event.schema.json" },
+    { "$ref": "../core/Reporter.schema.json" },
+    { "$ref": "../core/DateTime.schema.json" },
+    { "$ref": "../core/Statistics.schema.json" }
```
Vidushi-GitHub: How about adding localization?
There can be snews alert with RA, Dec values (https://gcn.gsfc.nasa.gov/other/150.snews)

gcn/notices/snews2/coincidence.test_alert.example.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/snews2/coincidence.schema.json",
+  "alert_datetime": "2026-04-03T16:23:21.937Z",
+  "alert_tense": "test",
+  "alert_type": "initial",
```
jracusin, line 5: Will you ever issue anything beyond "initial"? If so, add `record_number` from the Reporter core schema, where it increments as you provide updates.

jracusin:
In the [GCN Classic notices](https://gcn.gsfc.nasa.gov/snews_trans.html), you also had the number of neutrinos detected and event duration.  It would be interesting to have the number of neutrinos per detector.

gcn/notices/snews2/coincidence.test_alert.example.json
jracusin: Elsewhere in GCN Circulars and Notices, we just add "Z" rather than "+00:00". We prefer to be consistent.

Vidushi-GitHub:
Additionally, run `npx prettier -c .`
One of the validation is failing.

mlinvill:
> In the [GCN Classic notices](https://gcn.gsfc.nasa.gov/snews_trans.html), you also had the number of neutrinos detected and event duration. It would be interesting to have the number of neutrinos per detector.

We don't have neutrino counts in the coincidence tier. This information is reported in the timing tier, which is out of the scope of this initial effort. I expect our GCN schema will be extended to handle this information in future efforts.

gcn/notices/snews2/coincidence.test_alert.example.json
medhansh29: Thank you for the suggestion!
We have integrated your suggestion to track this sequence. SNEWS tracks these coincidence updates internally using a sub_list_number field. We have updated our transformation schema to automatically map this internal sequence tracker into the GCN core schemas.
We now populate the record_number field within the Reporter block to monotonically track these updates.

gcn/notices/snews2/coincidence.schema.json
medhansh29: We appended { "$ref": "../core/Localization.schema.json" } to the allOf array in coincidence.schema.json. However, rather than passing specific ra and dec coordinates, SNEWS will pass localization maps. To support this, we are utilizing the native healpix_url field provided by the core Localization schema to pass the URI for our skymaps.

medhansh29:
> Is there a formal scheme for naming these SNEWS events with an `event_name`?

No there is no formal event name, just the uuid passed generated which is passed into the $id field in the schema.

medhansh29:
> Additionally, run `npx prettier -c .` One of the validation is failing.

Fixed in latest update

jracusin:
There is an example for a test_alert, but can you please add a least one example for a real alert?  The schema includes localization, but there's no localization in the example. Would there be in a real alert? If it depends on which detectors are included, you could include multiple examples.

medhansh29:
> There is an example for a test_alert, but can you please add a least one example for a real alert? The schema includes localization, but there's no localization in the example. Would there be in a real alert? If it depends on which detectors are included, you could include multiple examples.

Two new examples have been added. One in which triangulation was achieved due to three or more participating detectors, and one in which triangulation was not achieved. We inherit localization, but we only use the Healpix map key from it since the new CoincidenceTier doesn't supply individual Ra and Dec values

gcn/notices/snews2/coincidence.real_alert_no_triangulation.example.json
```diff
@@ -0,0 +1,24 @@
+{
+  "$schema": "https://gcn.nasa.gov/schema/main/gcn/notices/snews2/coincidence.schema.json",
+  "alert_datetime": "2026-06-18T16:23:21.937Z",
+  "alert_tense": "current",
+  "alert_type": "initial",
+  "id": "SNEWS_Coincidence_ALERT 2026-06-18T16:23:21.937Z",
```
Vidushi-GitHub, line 6: Is trigger time is ID? 
Shouldn't it be detector observation ID?

jracusin: In GCN Classic, you had a trigger number, e.g. "TRIGGER_NUMBER:   SK_SN  10088". Is this still used?

Vidushi-GitHub:
@medhansh29 I have added information for new Notices on SNEWS mission page, PR (https://github.com/nasa-gcn/gcn.nasa.gov/pull/3729).
Please review the PR.

Issue #388, opened 2026-07-24
Publishing to NPM is broken


lpsinger:
The last version successfully published to NPM was version 6.1.1 on February 3, 2026.

lpsinger:
@dakota002, would you or someone else please do a release of this package to publish it to NPM?

dakota002:
> [@dakota002](https://github.com/dakota002), would you or someone else please do a release of this package to publish it to NPM?

Sure, I can trigger it in a sec

lpsinger:
That worked!

PR #391, opened 2026-07-27
Fix publish workflow permissions

lpsinger:
See #388.


dakota002 (approved):

lpsinger:
Would you please try publishing another release?

dakota002:
Would this repo benefit from adding the [workflow_dispatch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch) trigger so we can just click a button and run the workflow?

PR #401, opened 2026-08-25
modify example file names for schema browser rules

Vidushi-GitHub:
# Description
SVOM examples are not available in the schema browser, as don't follow the `gcn/notices/mission/schema_name.optional_description.example.json` naming convention.

The files are removed from folder and renamed as per rule.


dakota002 (changes requested):
This puts them in the correct location, but the `optional_description` part should be using only underscores. For example:
`trigger.eclairs.catalog.example.json` should be `trigger.eclairs_catalog.example.json`

dakota002 (approved):

jracusin (approved):

